PR #5, opened 2023-10-12
Update to latest fluree/db with nameservice abstraction

bplatz:
Update to use the latest fluree/db, with modifications needed to support the new nameservice abstraction that was put in this PR: https://github.com/fluree/db/pull/590


zonotope (approved):
🦞

PR #6, opened 2023-10-19
Fix Docker Build/Run Issues

aaj3f:
A few changes here

- Dockerfile used wrong artifact names so that the `ENTRYPOINT` failed on `docker run`
- `config.edn` needed a `:docker` profile value for `consensus-this-server`
- (it looks like my Makefile suggestions for `docker-build`, `docker-run`, and `docker-push` made their way in here somehow!) `docker-build` needed a `--load` flag to make it into a local machine's image repository


zonotope (approved):
⛵

PR #8, opened 2023-10-24
Update transact/create handling to use new insert syntax

dpetran:
This updates server to use the transaction syntax.

Depends upon https://github.com/fluree/json-ld/pull/27

### DIFF
--- a/src/fluree/server/consensus/handlers/create_ledger.clj
+++ b/src/fluree/server/consensus/handlers/create_ledger.clj
@@ -1,10 +1,11 @@
 (ns fluree.server.consensus.handlers.create-ledger
   (:require [clojure.core.async :as async]
+            [fluree.db.constants :as const]
             [fluree.db.json-ld.api :as fluree]
             [fluree.db.util.json :as json]
+            [fluree.db.util.log :as log]
             [fluree.raft.leader :refer [is-leader?]]
             [fluree.server.handlers.shared :refer [deref!]]
-            [fluree.db.util.log :as log]
             [fluree.server.consensus.core :as consensus]
             [fluree.server.consensus.producers.new-index-file :refer [push-new-index-files]]
             [fluree.server.consensus.raft.core :as raft]))
@@ -24,12 +25,21 @@
         false)
       true)))
 
+(defn parse-opts
+  "Extract the opts from the transaction and keywordify the top level keys."
+  [expanded-txn]
+  (-> expanded-txn
+      (get const/iri-opts)
+      (get 0)
+      (get "@value")
+      (->> (reduce-kv (fn [opts k v] (assoc opts (keyword k) v)) {}))))
+
 (defn create-and-commit
   "Returns promise with eventual response, which could be an exception."
   [{:keys [fluree/conn] :as config}
-   {:keys [ledger-id txn opts tx-id] :as _params}]
+   {:keys [ledger-id txn tx-id] :as _params}]
   (log/trace "Creating ledger" ledger-id "with txn:" txn)
-  (let [txn'    (json/parse txn false)
+  (let [opts    (parse-opts txn)
         ledger  (deref! (fluree/create conn ledger-id opts))
         commit! (fn [db]
                   (let [index-files-ch (async/chan)
@@ -41,10 +51,9 @@
     ;; following uses :file-data? and will return map with {:keys [db data-file commit-file]}
     (-> ledger
         fluree/db
-        (fluree/stage txn' opts)
+        (fluree/stage2 txn)
         deref!
         commit!)))
-
 
 (defn push-consensus
   "Pushes create-ledger request in consensus only if leader.
--- a/src/fluree/server/consensus/handlers/tx_queue.clj
+++ b/src/fluree/server/consensus/handlers/tx_queue.clj
@@ -92,8 +92,8 @@
         commit! (fn [db]
                   (let [index-files-ch (async/chan)
                         ;; monitor for new index files and push across network
-                        _ (push-new-index-files config index-files-ch)
-                        resp (fluree/commit! ledger db {:file-data? true
+                        _    (push-new-index-files config index-files-ch)
+                        resp (fluree/commit! ledger db {:file-data?     true
                                                         :index-files-ch index-files-ch})]
                     (log/debug "New commit for ledger" ledger-id "with tx-id: " tx-id
                                "processed in" (- (System/currentTimeMillis) start-time) "milliseconds.")
--- a/src/fluree/server/consensus/handlers/tx_queue.clj
+++ b/src/fluree/server/consensus/handlers/tx_queue.clj
@@ -1,5 +1,6 @@
 (ns fluree.server.consensus.handlers.tx-queue
   (:require [clojure.core.async :as async]
+            [fluree.db.constants :as const]
             [fluree.db.json-ld.api :as fluree]
             [fluree.db.util.core :as util]
             [fluree.db.util.json :as json]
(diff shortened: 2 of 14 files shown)


src/fluree/server/consensus/handlers/create_ledger.clj
```diff
@@ -24,12 +25,21 @@
         false)
       true)))
 
+(defn parse-opts
+  "Extract the opts from the transaction and keywordify the top level keys."
+  [expanded-txn]
+  (-> expanded-txn
+      (get const/iri-opts)
+      (get 0)
```
cap10morgan, line 33: Equivalent to `first`?

src/fluree/server/consensus/handlers/tx_queue.clj
```diff
@@ -100,27 +78,30 @@
...
-                       resp))]
+
+        default-context (-> txn (get const/iri-default-context) (get 0) (get "@value"))
+
+        commit! (fn [db]
+                  (let [index-files-ch (async/chan)
+                        ;; monitor for new index files and push across network
+                        _ (push-new-index-files config index-files-ch)
+                        resp (fluree/commit! ledger db {:file-data? true
+                                                        :index-files-ch index-files-ch})]
```
cap10morgan, line 97: Can you re-align those let bindings?

src/fluree/server/consensus/handlers/tx_queue.clj
```diff
@@ -100,27 +78,30 @@
...
+                    (log/debug "New commit for ledger" ledger-id "with tx-id: " tx-id
+                               "processed in" (- (System/currentTimeMillis) start-time) "milliseconds.")
+                    resp))]
     (-> ledger
         fluree/db
-        (update-default-context defaultContext)
-        (fluree/stage txn' opts)
+        (update-default-context default-context)
```
cap10morgan, line 103: image

src/fluree/server/handlers/create.clj
```diff
@@ -29,15 +30,13 @@
           (watcher/deliver-watch watcher tx-id persist-resp))))))
 
 (defn create-ledger
-  [p consensus conn watcher parsed-txn]
-  (let [{ledger-id const/iri-ledger
-         txn       "@graph"
-         opts      :opts} parsed-txn
-        [tx-id txn*] (normalize-txn txn)
+  [p consensus conn watcher expanded-txn]
+  (let [ledger-id     (-> expanded-txn (get const/iri-ledger) (get 0) (get "@value"))
```
cap10morgan, line 34: Starting to feel like we could use a `get-expanded-value` (or similar) fn along the lines of:

```clojure
(defn get-expanded-value
  [expanded iri-key]
  (-> expanded (get iri-key) first (get "@value"))
```

src/fluree/server/handlers/create.clj
```diff
@@ -76,21 +75,12 @@
...
-         default-context const/iri-default-context
-         :as parsed-txn}
-        (transact-api/parse-json-ld-txn conn context-type body)
-        opts*        (cond-> (or (:opts body) {})
-                       txn-opts (merge txn-opts)
-                       txn-context (assoc :txn-context txn-context)
-                       default-context (assoc :defaultContext default-context))]
+  (let [[expanded-txn] (util/sequential (jld-processor/expand body))
+        ledger-id  (-> expanded-txn (get const/iri-ledger) (get 0) (get "@value"))
+        resp-p       (promise)]
```
cap10morgan, line 80: Needs alignment

src/fluree/server/handlers/transact.clj
```diff
@@ -20,22 +20,16 @@
...
+(defn derive-tx-id
   [txn]
-  (let [txn*  (json-ld/normalize-data
-                txn
-                {:algorithm :basic
-                 :format    :application/json})
-        tx-id (hash-txn txn*)]
-    [tx-id txn*]))
+  (let [canonized  (jld-processor/canonize txn)]
+    (hash-txn canonized)))
```
cap10morgan, line 26: Could be simplified to `(-> txn jld-processor/canonize hash-txn)`

src/fluree/server/handlers/transact.clj
```diff
@@ -20,22 +20,16 @@
...
 
 (defn queue-consensus
-  [consensus conn watcher ledger tx-id txn* default-context opts]
+  [consensus conn watcher ledger tx-id expanded-txn]
   (let [conn-type       (conn-proto/-method conn)
         ;; initial response is not completion, but acknowledgement of persistence
-        persist-resp-ch (consensus/queue-new-transaction
-                         consensus conn-type ledger tx-id txn* default-context
-                         opts)]
+        persist-resp-ch (consensus/queue-new-transaction consensus conn-type ledger tx-id expanded-txn nil nil)]
```
cap10morgan, line 32: Can you split this up into two lines? (My tall, narrow editor windows thank you in advance!)

test/fluree/server/integration/basic_query_test.clj
```diff
@@ -11,8 +11,9 @@
     (let [ledger-name (create-rand-ledger "query-endpoint-basic-entity-test")
           txn-req     {:body
                        (json/write-value-as-string
-                         {"f:ledger" ledger-name
-                          "@graph"   [{"id"      "ex:query-test"
+                         {"ledger" ledger-name
```
cap10morgan, line 14: alignment (and others below)

test/fluree/server/integration/basic_transaction_test.clj
```diff
@@ -11,9 +11,9 @@
   (testing "can create a new ledger w/ JSON"
     (let [ledger-name (str "create-endpoint-" (random-uuid))
           req         (json/write-value-as-string
-                        {"f:ledger" ledger-name
-                         "@context" ["" {"foo" "http://foobar.com/"}]
-                         "@graph"   [{"id"      "ex:create-test"
+                        {"ledger" ledger-name
+                         "@context" ["https://ns.flur.ee" {"foo" "http://foobar.com/"}]
```
cap10morgan, line 15: I think this should be in the conn default context if we're going to need it in every transaction. I think it would make more sense to keep its `f` alias in there, though, to keep that namespace cleaner.

cap10morgan (approved):
Some minor issues & suggestions. Looks good overall.

src/fluree/server/handlers/create.clj
dpetran: Since this is dealing with `jld-processor/expand` output and not `json-ld/expand` output, I thought it would be confusing to have two virtually identical functions that won't work if you pick the wrong one. I know the manual unthreading is verbose but it's very readable and nobody will get bit by using the wrong helper down the road. I can still try to abstract if you like, but in this case I didn't think the abstraction was worth the cost.

dpetran:
Rebased on main to resolve conflicts.

src/fluree/server/handlers/create.clj
cap10morgan: Hmm... not sure I follow. I saw the exact same sequence of fns applied in several places and suggested turning that into a fn that gets called instead. But maybe I missed the places that were slightly different? Proper namespacing & docstring might be good if you think it might o/w look tempting to misuse. But yeah, I think it makes sense to turn into a named fn. I'll leave the final decision for this PR up to you, though.

test/fluree/server/integration/basic_transaction_test.clj
dpetran: I've updated the default conn default context to include it, but users will still need to specify at least the `https://ns.flur.ee` as an explicit `@context` value in every transaction, at least for now. That way credentials will work and we can always expand the transaction without having to check the conn default context.

I do think there's a path towards not including it, it just isn't implemented yet.

PR #10, opened 2023-10-24
fix default conn default context

dpetran:
Downstream code expects the conn default context to be a map, not a vector. This backs out the change.


cap10morgan (approved):
🛠️

PR #12, opened 2023-10-25
use query error formatting fn in query error coercion

mpoffald:
Related to https://github.com/fluree/db/issues/312

Uses error-formatting fn introduced in https://github.com/fluree/db/pull/598 to format validation errors for query endpoints


mpoffald:
Closing in favor of https://github.com/fluree/server/pull/13, which contains the same changes

PR #13, opened 2023-10-27
Feature/subscription

bplatz:
This is companion support for `fluree/db` subscriptions: https://github.com/fluree/db/pull/605

When connecting to a fluree/server, the fluree/db library now launches a socket connection to keep up to date on ledger changes and maintains its internal state and a current copy of db's that are actively being used from the library.

### DIFF
--- a/src/fluree/server/components/subscriptions.clj
+++ b/src/fluree/server/components/subscriptions.clj
@@ -1,0 +1,219 @@
+(ns fluree.server.components.subscriptions
+  (:require [clojure.core.async :as async]
+            [donut.system :as ds]
+            [fluree.db.util.core :as util]
+            [fluree.db.util.json :as json]
+            [fluree.db.util.log :as log]
+            [ring.adapter.jetty9.websocket :as ws])
+  (:import (java.io IOException)
+           (java.nio.channels ClosedChannelException)))
+
+
+;; structure of subscriptions atom:
+;{:subs    {"id" {:chan   port ;; <- core.async channel
+;                 :ledgers {"some/ledger" {}} ;; <- map where each key is a ledger-id subscribed to
+;                 :did     nil}}
+; :closed? false}
+
+(def default-subscription-atom (atom {:subs    {}
+                                      :closed? false}))
+
+(defn close-chan
+  [chan message]
+  ;; events are of structure [ledger-id event-name event-data]
+  (when message
+    (async/put! chan [nil :closing message]))
+  (async/close! chan))
+
+(defn all-chans
+  [sub-state]
+  (->> sub-state
+       :subs
+       vals
+       (map :chan)))
+
+(defn all-sub-ids
+  [sub-state]
+  (->> sub-state
+       :subs
+       keys))
+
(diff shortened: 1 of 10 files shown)


src/fluree/server/components/subscriptions.clj
```diff
@@ -0,0 +1,219 @@
...
+        "ping" ;; Note in-browser websocket doesn't support ping/pong, so this is a no-op to keep alive
+        (do
+          (log/trace "Websocket keep-alive from sub-id:" sub-id)
+          nil)))
+    (catch Exception e
+      (log/error e "Error with :on-text message from websocket subscriber: " sub-id))))
+
+(defmethod client-message :on-bytes
+  [{:keys [http/sub-id payload offset len]}]
+  (log/info "websocket :on-bytes (no-op) message for sub-id: " sub-id))
```
cap10morgan, line 206: The log fns will insert the spaces between args for you (like `println`)

src/fluree/server/components/subscriptions.clj
```diff
@@ -0,0 +1,219 @@
...
+
+;; pong handled automatically, no response needed
+(defmethod client-message :on-ping
+  [{:keys [http/sub-id payload]}]
+  (log/trace "ws :on-ping message received from: " sub-id "with payload: " payload))
+
+(defmethod client-message :on-error
+  [{:keys [http/sub-id error]}]
+  (log/warn "Websocket error for sub-id: " sub-id "with error: " (type error)))
```
cap10morgan, line 215: Should we log this at `error` level? Or are they so common that `warn` is better?

src/fluree/server/components/subscriptions.clj
```diff
@@ -0,0 +1,219 @@
...
+        (when (contains? ledgers ledger-id)
+          ;; TODO - use ledger subscription opts to filter out messages, permission
+          (send-message subscriptions sub-id message))))))
+
+
+(defn send-message-to-all
+  "Sends a message to all subscriptions. Message sent as JSON stringified map
+  in the form: {action: commit, ledger: my/ledger, data: {...}}"
+  [{:keys [sub-atom] :as subscriptions} action ledger-alias data]
+  (log/warn "SEND MESSAGE TO ALL subscriptions: " subscriptions)
```
cap10morgan, line 150: Why is this logged at `warn` level?

cap10morgan (approved):
A few small logging things to consider / clean up. But very minor and looks good overall!

src/fluree/server/components/subscriptions.clj
bplatz: Very common, just a network connection break. There is a list of error types, and I'd suspect the optimal solution will be to eventually `case` the error type, with some being log/debug all the way up to log/error. So log/warn is sort of an imperfect compromise. I think we'll have a better sense of what is best here once it is running.

src/fluree/server/components/subscriptions.clj
bplatz: This was a debugging message, I'll remove.

PR #17, opened 2023-11-03
Add and update consensus configuration options

bplatz:
This primarily adds additional consensus configuration options and provides documentation for them.

The default logging directory now has some additional logic to str/replace the '/' in the multi-address format to '-' such that a deeply nested directory is not created.

### DIFF
--- a/src/fluree/server/consensus/core.clj
+++ b/src/fluree/server/consensus/core.clj
@@ -1,6 +1,7 @@
 (ns fluree.server.consensus.core
   (:require [clojure.core.async :as async]
             [clojure.string :as str]
+            [fluree.db.conn.proto :as conn-proto]
             [fluree.db.util.log :as log]
             [fluree.raft :as raft]
             [fluree.server.consensus.network.tcp :as ftcp]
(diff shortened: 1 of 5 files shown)


bplatz:
Being based on top of feature/where-syntax-update which still has the failing policy test. Once that passes this should pass.

src/fluree/server/consensus/core.clj
```diff
@@ -153,33 +154,66 @@
...
-                              io-file/canonicalize-path)
-        ledger-directory* (-> (or ledger-directory
-                                  (str "./data/" (name this-server) "/ledger/"))
-                              io-file/canonicalize-path)]
-    (assoc raft-config :log-directory log-directory*
-                       :ledger-directory* ledger-directory*)))
-
+  [{:keys [log-path conn-storage-path this-server] :as raft-config}]
+  (log/warn "LOG PATH: " log-path)
+  (log/warn "CONN STORAGE PATH: " conn-storage-path)
```
cap10morgan, line 178: Do we want these logged at the `warn` level? Or at all here?

src/fluree/server/consensus/core.clj
```diff
@@ -153,33 +154,66 @@
...
+  (let [log-path*          (-> (or (some-> log-path add-trailing-slash)
+                                   (str "./data/" (path-safe-server-address this-server) "/_log/"))
+                               io-file/canonicalize-path)
+        conn-storage-path* (-> (or conn-storage-path
+                                   (str "./data/" (path-safe-server-address this-server) "/"))
+                               io-file/canonicalize-path)]
+    (log/warn "LOG PATH*: " log-path*)
+    (log/warn "CONN STORAGE PATH*: " conn-storage-path*)
```
cap10morgan, line 186: Do we want to leave these log messages in?

src/fluree/server/consensus/core.clj
```diff
@@ -153,33 +154,66 @@
...
+    (assoc raft-config :log-path log-path*
+                       :log-directory log-path* ;; note raft library config is `:log-directory`
+                       :conn-storage-path conn-storage-path*)))
+
+
+(defn is-shared-storage?
+  "If conn method is :file or :memory, returns false, else true."
+  [conn]
+  (let [method (conn-proto/-method conn)]
+    (if (#{:file :memory} method)
```
cap10morgan, line 196: Can we reference this set from a shared ns? I see it's defined as `fluree.server.components.consensus/replicate-file-conn-types`, but not sure if we can access that here w/o a circular ref.

cap10morgan (approved):
A couple of log messages to maybe change / remove and some DRYing up to consider. Looks good once addressed in whatever way seems best.

bplatz:
This is quite old now and not really relevant so closing.

PR #18, opened 2023-11-03
Feature/where syntax update

bplatz:
This updates tests to use new `where` syntax on latest fluree/db.


bplatz:
There is a failing policy test related to the current fluree/db that needs to get addressed before this PR can go forward.

bplatz:
FYI, this branch was tied to feature/subscription which was failing for the same reason and already approved (and now merged)

PR #19, opened 2023-11-14
update fluree api usage

dpetran:
This propagates the api changes from https://github.com/fluree/db/pull/679 to the server.


bplatz (approved):
Yay!

PR #24, opened 2023-12-14
Fix/remove default context

zonotope:
This patch updates server not to rely on default contexts in connections.


dpetran (approved):
:deciduous_tree:

PR #25, opened 2023-12-15
Add cljfmt config

cap10morgan:

PR #26, opened 2023-12-15
Run cljfmt fix src test build.clj

cap10morgan:


cap10morgan:
Current dependencies on/for this PR:
* `main`
  * **PR #25** <a href="https://app.graphite.dev/github/pr/fluree/server/25?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #26** <a href="https://app.graphite.dev/github/pr/fluree/server/26?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/fluree/server/26?utm_source=stack-comment).

dpetran (approved):
:champagne:

src/fluree/server/components/consensus.clj
```diff
@@ -11,10 +11,10 @@
   (log/debug "Starting raft consensus with config:" config)
   (let [handler (consensus-handler/create-handler consensus-handler/default-routes)]
     (consensus/start handler (assoc config :join? false
```
zonotope, line 13: i'm not sure if cljfmt can handle this, but i think multi-line assoc calls should have their first key-value pair start on a new line so that all the k-v pairs line up for easy scanning. here, the `:join? false` is on a different column from all the other pairs, so someone doing a quick scan to see which attributes are set to one could easily miss it.

src/fluree/server/consensus/core.clj
```diff
@@ -66,36 +66,36 @@
    :state-atom     state-machine-atom
    :storage-read   (or storage-group-read
                        (io-file/connection-storage-read
-                         log-directory
-                         encryption-key))
+                        log-directory
```
zonotope, line 69: i think i prefer 2-space indentation for function calls when all the arguments are on a new line. i think it better expresses the semantic difference between the function and the arguments it takes. but i'm also ok with leaving it this way if other folks disagree.

src/fluree/server/consensus/network/multi_addr.clj
```diff
@@ -32,39 +32,39 @@
...
-                            " Port value must be an integer. "
-                            "Supplied: " port))))]
+                      (str "Error in multi-address port in: " multi-addr
+                           " Port value must be an integer. "
+                           "Supplied: " port))))]
     (when-not ({"ip4" "dns4" "ip6" "dns6"} address-type)
       (invalid-address-shutdown
-        (str "Error in multi-address type in: " multi-addr
-             " Fluree only supports the ipv4, ipv6, dns4 or dns6 at the moment.")))
+       (str "Error in multi-address type in: " multi-addr
```
zonotope, line 45: i think this is an even better example of why 2 space indentation for function calls is more legible.

src/fluree/server/components/http.clj
```diff
@@ -1,30 +1,30 @@
 (ns fluree.server.components.http
   (:require
```
zonotope, line 2: Stuart Sierra notwithstanding, I prefer the first ns require to be on the same line as the `:require` keyword. I think the extra indentation that gives helps break up the require blocks vs imports, uses, and refer-clojures.

src/fluree/server/components/consensus.clj
cap10morgan: Yeah, I don't _think_ it can enforce that.

src/fluree/server/consensus/core.clj
cap10morgan: This is actually just cljfmt treating lists like other multiline compound data literals. For example:

```clojure
[:foo
 :bar
 :baz]
```

I'm not sure if you can tell it to make an exception for lists or if you'd have to make all such data literals indent w/ two spaces on subsequent lines.

src/fluree/server/components/http.clj
cap10morgan: I tend to agree, but I think this is another thing that cljfmt cannot (yet?) enforce.

src/fluree/server/consensus/core.clj
cap10morgan: Looks like I was wrong and this is specifically configurable in cljfmt: https://github.com/weavejester/cljfmt#formatting-options (see `:function-arguments-indentation` section). I'm happy to go w/ whatever the team consensus is here.

dpetran: I like 2-space indentation here as well

cap10morgan: Do you all have a preference between the `:cursive` and `:zprint` options for that param?

cap10morgan:
### Merge activity

* **Jan 2, 2:26 PM**: @cap10morgan started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/fluree/server/26).

PR #32, opened 2024-01-16
Add optional git pre-commit hooks for cljfmt & clj-kondo

cap10morgan:
This uses babashka tasks to run some pre-commit checks via git hooks, but only after you opt in by running `bb run git-hooks install`. Currently it runs cljfmt and clj-kondo and rejects the commit if either complain. In my testing this runs in tens of milliseconds so hopefully worth the added value to each commit.

Currently it only checks files that are staged, and crucially, it checks the staged version of those files in a separate tmp dir (every example of git hooks I saw online checked the working directory version of the staged files; which are often not the same!). If we find a need to have the other un-staged files in there too, I can add that. I wasn't sure if it made a difference or not.

It is currently depending on a fork of cljfmt because it didn't have a library fn to pass in paths to recursively check. I'm going to PR that upstream and see if I can get it or something similar into an official release.

This is all towards the goal of spending less time fixing bugs that sneak onto `main` and wasting time cleaning up conflicts that are merely due to inconsistent code formatting. We had been doing far too much of the latter, especially.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,10 @@ The most common environment variables to set would include:
 - If a server fails, it is auto-detected and the work of the server is redistributed to the remaining servers in the cluster
 - If a server joins, work is redistributed to include the new server
 
+## Development
+
+### Git hooks
+
+- Run `bb git-hooks install`
+    - This will set up a pre-commit git hook that checks your staged changes
+      with cljfmt and clj-kondo.
--- a/bb.edn
+++ b/bb.edn
@@ -1,0 +1,6 @@
+{:paths ["bb"]
+ :pods  {clj-kondo/clj-kondo {:version "2023.12.15"}}
+ :deps  {dev.weavejester/cljfmt {:git/url "https://github.com/cap10morgan/cljfmt.git"
+                                 :git/sha "acc6a7d1e5a7e3391419ac731bf8a2774b8c4a13"}}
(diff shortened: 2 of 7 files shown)


cap10morgan:
Current dependencies on/for this PR:
* `main`
  * **PR #32** <a href="https://app.graphite.dev/github/pr/fluree/server/32?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/fluree/server/32?utm_source=stack-comment).

README.md
```diff
@@ -78,3 +78,10 @@ The most common environment variables to set would include:
 - If a server fails, it is auto-detected and the work of the server is redistributed to the remaining servers in the cluster
 - If a server joins, work is redistributed to include the new server
 
+## Development
+
+### Git hooks
+
+- Run `bb git-hooks install`
```
dpetran, line 85: This is what I'd need to do for this to work going forwards, right?

dpetran (approved):
:+1:

README.md
cap10morgan: Yep, or perhaps more properly `bb run git-hooks install` but this should work too.

PR #36, opened 2024-01-31
add test for successful jws tests

dpetran:
The request body needs to be json encoded if it's sent with json headers, just a string will cause weird/broken behavior.

### DIFF
--- a/test/fluree/server/integration/policy_test.clj
+++ b/test/fluree/server/integration/policy_test.clj
@@ -147,12 +147,25 @@
                   "policy opts prevented seeing bob's secret"))))
         (testing "JWS requests"
           (let [txn-req {"@context" ["https://ns.flur.ee" test-system/default-context]
+                         "ledger" ledger-name
+                         "insert" {"id" "ex:alice"
+                                   "ex:secret" "Alice's new secret"}}
+                txn-res (api-post :transact {:body (json/write-value-as-string
+                                                    (crypto/create-jws
+                                                     (json/write-value-as-string txn-req)
+                                                     (:private auth)))
+                                             :headers json-headers})]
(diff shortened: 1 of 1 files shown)


test/fluree/server/integration/policy_test.clj
```diff
@@ -146,13 +146,26 @@
                      (-> query-res :body json/read-value first (get "f:assert")))
                   "policy opts prevented seeing bob's secret"))))
         (testing "JWS requests"
+          (let [txn-req {"@context" ["https://ns.flur.ee" test-system/default-context]
```
zonotope, line 149: Could you add more commentary to these tests with more `testing` macro calls?

What is this setup code before the test assertion doing? Why should this transaction be authorized? Why should the following one not be?

dpetran: This whole deftest block sets up a simple graph which allows users to view and update their own `ex:secret` predicates. The jws block of tests verifies that using the identity can be correctly verified and used for policy enforcement via the jws mechanism of signing requests.

Before, there were only two cases: that Alice couldn't modify Bob's secret and that Bob's secrets don't appear in Alice's query results. I added the test case for Alice being able to modify her own secret.

zonotope (approved):
I think the test documentation here could be better, but that would require overhauling the whole file, which is beyond the scope of this pr.

I do think we need to do a better job of our test documentation in general however. I think the `testing` and `is` macros should be set up to provide a full English sentence describing the whole setup of each test and what should happen on each assertion. This will make it easier for others who did not write the code or tests to understand what the intentions are. 

I think his is something we should keep in mind moving forward, but this specific pr is 👍🏾

PR #39, opened 2024-02-14
add tests to verify list deletion

dpetran:
This is to verify that https://github.com/fluree/core/issues/74#issuecomment-1944557992 is fixed.


cap10morgan (approved):
❌

PR #40, opened 2024-02-17
Feature/sync sid calculation

zonotope:
This patch updates server work with the db library using SID objects introduced in fluree/db#660. The api is very similar on the db side, so the only things that had to change here are (1) removing the negation of t-values in commit responses because those values are now always positive, and (2) removing the expectation that the "id" attribute will exist after a deletion because we no longer have "id" flakes in the db.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@
 /.clj-kondo/*
 !/.clj-kondo/config.edn
 /.lsp/
-/.dir-locals.el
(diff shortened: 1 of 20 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@
 /.clj-kondo/*
 !/.clj-kondo/config.edn
 /.lsp/
+/.dir-locals.el
```
cap10morgan, line 12: Would be better to put this in your global git ignore file in your home dir.

cap10morgan (approved):
One nit-pick-y suggestion on the .gitignore addition, but LGTM o/w!

PR #41, opened 2024-02-28
Task/adopt storage proto

dpetran:
Also did a small update to accommodate the new sid work.


cap10morgan:
CI failure is due to: `Error building classpath. The following libs must be prepared before use: [com.fluree/db]`

PR #43, opened 2024-03-12
Add `:remote` profile configuration

flureeadmin:
In addition to adding a `:remote` profile to `config.edn` and an alias for running as `:remote` to `deps.edn`, also removed stale config for a `DEFAULT CONTEXT`

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -52,6 +52,14 @@
   :run-dev
   {:exec-fn   fluree.server.main/run-server
    :exec-args {:profile :dev}}
+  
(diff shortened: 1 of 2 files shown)


cap10morgan:
I was trying to decide if a config profile was the right way to do this, but I couldn't convince myself it wasn't. So 👍🏻

cap10morgan (approved):

bplatz:
We no longer config like this, and this PR is outdated so closing.

PR #48, opened 2024-04-01
Refactor Consensus Protocol

zonotope:
This patch among other things is a fix for the raft command queue being overwhelmed during a large reindexing process because of a lack of back pressure. The changes adding back pressure when forwarding new index file notifications to the consensus group are all contained in [this commit](https://github.com/fluree/server/commit/6ca05f979a55c4a13d5dec277957bb9f0b8e3a88)

In the process of identifying changes we could make to the consensus api I found that, even though we had a `TxGroup` protocol for consensus, we had raft-specific implementation details throughout the code on either side of the protocol, so the rest of these changes are meant to isolate the raft implementation and tighten up the consensus protocol to make it more modular, extensible, and easier to add new consensus types in the future. I have [refactored the `TxGroup` protocol to only contain the methods we need for general consensus](https://github.com/fluree/server/blob/863ba05dea6f7d5d99255dde048bc7a090889b2d/src/fluree/server/consensus.clj#L8-L10), moved all raft-specific implementation details to the `fluree.server.consensus.raft*` namespace hierarchy, and removed unused code.


cap10morgan:
@zonotope Looks like a potentially legit test failure? I think the expectation just needs to be updated (or the test itself to have something reasonable to expect now) due to the changes brought in by the db version bump.

zonotope:
> @zonotope Looks like a potentially legit test failure? I think the expectation just needs to be updated (or the test itself to have something reasonable to expect now) due to the changes brought in by the db version bump.

yep, a legit test failure brought on by the db version bump. I'm pretty sure it came from https://github.com/fluree/db/pull/744. I'm still thinking about the best way to fix it though, and the answer might be more philosophical than technical. if you query for an iri specifically, but we have no facts about it, should we give you nothing or an id map with that iri and nothing else? I could make the argument either way.

zonotope:
> > @zonotope Looks like a potentially legit test failure? I think the expectation just needs to be updated (or the test itself to have something reasonable to expect now) due to the changes brought in by the db version bump.
> 
> yep, a legit test failure brought on by the db version bump. I'm pretty sure it came from [fluree/db#744](https://github.com/fluree/db/pull/744). I'm still thinking about the best way to fix it though, and the answer might be more philosophical than technical. if you query for an iri specifically, but we have no facts about it, should we give you nothing or an id map with that iri and nothing else? I could make the argument either way.

I went with the id map with nothing else instead of empty results. The subject exists by virtue of the user specifying the iri; we just don't know anything about it. The id map fits most closely with that I think.

src/fluree/server/consensus/raft/participant.clj
```diff
@@ -0,0 +1,55 @@
...
+         raft-stub {:config {:command-chan command-chan}}]
+     (raft/new-entry raft-stub entry callback (or timeout-ms 5000)))))
+
+(defn leader-new-command!
+  "Issue a new command but only if currently the leader, else returns an exception.
+
+  This is used in the consensus handler to issue additional raft commands after doing
+  some work.
+
+  Returns a promise channel, which could contain an exception so be sure to check!"
```
cap10morgan, line 31: Remove the word `channel` now? Looks like it just returns a regular ol' promise.

src/fluree/server/consensus/raft/participant.clj
```diff
@@ -0,0 +1,55 @@
...
+
+  ([config command params timeout-ms]
+   (assert-leader config command)
+   (let [p         (promise)
+         callback  (fn [resp]
+                     (deliver p resp))]
+     (new-command! config command params timeout-ms callback)
+     p)))
+
+(defn leader-new-command!-async
```
cap10morgan, line 43: Might be good to add a docstring here

src/fluree/server/consensus/raft/participant.clj
```diff
@@ -0,0 +1,55 @@
...
+  ([config command params]
+   (leader-new-command!-async config command params 5000))
+
+  ([config command params timeout-ms]
+   (assert-leader config command)
+   (let [ch       (async/chan)
+         callback (fn [resp]
+                    (async/put! ch resp
+                                (fn [_]
+                                  (async/close! ch))))]
```
cap10morgan, line 53: Any reason not to use a `promise-chan` for this?

src/fluree/server/consensus/raft/producers/new_index_file.clj
```diff
@@ -34,8 +33,8 @@
           (log/error next-file-event "Error in push-new-index-files monitoring new index files, but got an exception.")
           (do
             (case (:event next-file-event)
-              :new-index-file (push-index-file config next-file-event)
-              :new-commit (consensus-push-index-commit config (:data next-file-event)))
+              :new-index-file (<! (push-index-file config next-file-event))
```
cap10morgan, line 36: 💙

src/fluree/server/consensus/watcher.clj
```diff
@@ -0,0 +1,61 @@
+(ns fluree.server.consensus.watcher
+  "When mutation requests happen over http (e.g. creating a new ledger), we want
+  to wait for the operation to complete through consensus before providing the
+  response back to the API requester. These operations get delivered via
+  consensus, where they end up being picked up by a responsible server from a
+  queue, processed, and then the results broadcast back out."
```
cap10morgan, line 6: I like ns docstrings and I think we should use them more often, but this one kind of feels like it's missing the part about what this ns does. I think it's good to keep what's here as an example use case, but would love to have it start with a short explanation of what this ns is for in the abstract.

src/fluree/server/consensus/watcher.clj
```diff
@@ -0,0 +1,61 @@
...
+  [{:keys [max-tx-wait-ms watcher-atom]
+    :or   {max-tx-wait-ms 60000}}]
+  {:max-tx-wait-ms max-tx-wait-ms
+   :watcher-atom   (or watcher-atom
+                       (new-watcher-atom))})
+
+(defn close
+  [watcher]
+  (let [watcher-atom (:watcher-atom watcher)
+        watches (vals @watcher-atom)]
```
cap10morgan, line 59: let form alignment

cap10morgan (approved):
Some comments, suggestions, and questions to consider. Once you're happy I'm happy. Overall, looks like a great improvement!

src/fluree/server/consensus/raft/participant.clj
zonotope: Nah, I think that should work

src/fluree/server/consensus/watcher.clj
zonotope: That makes sense. This was a raw comment at the top of the ns and I just copied it verbatim into the ns docstring instead. I'll try to add a description of what this ns does as well.

PR #49, opened 2024-04-07
feat: Add user fluree to run the container as non-root

jrmanes:
hello!

I've added the user `fluree` (10001) to run the container as non-root.

Tested and working:
```bash
... [async-dispatch-6] INFO  fluree.server.consensus.raft.core - Ledger group leader change: {:cause :become-leader, :server "/ip4/127.0.0.1/tcp/62071", :new-leader "/ip4/127.0.0.1/tcp/62071", :old-leader nil, :event :become-leader, :message "This server, /ip4/127.0.0.1/tcp/62071, received the majority of the votes to become leader. New term: 1, latest index: 0."}
Fluree HTTP API server running on port 8090
```

cheers! 🚀

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,26 @@ RUN clojure -T:build uber
 
 FROM eclipse-temurin:17-jre-jammy AS runner
 
-RUN mkdir -p /opt/fluree-server
-WORKDIR /opt/fluree-server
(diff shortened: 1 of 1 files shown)


cap10morgan (approved):
LGTM

Thanks @jrmanes!

PR #52, opened 2024-04-16
require application/jwt content type for jwt requests

dpetran:
You can still json-encode a JWT like before, but using the application/jwt content-type for requests will be the documented way forward.

Ran into some trouble with the history coercion/validation/transformation, and I think I've got it sorted by adding the `fql-transformer` to the coercer.

Fixes https://github.com/fluree/core/issues/92


zonotope (approved):
🛷

PR #56, opened 2024-04-24
support query-string t val in from, from-named

bplatz:
Incorporates the new query-string t value syntax in fluree/db as per this PR:
https://github.com/fluree/db/pull/761

The functionality is exposed through the `/query` API allowing query `from` and `from-named` keys to specify t values as per the following example:

- my/db?t=42
- my/db?t=2020-01-01T00:00:00Z

Adds a test

### DIFF
--- a/test/fluree/server/integration/basic_transaction_test.clj
+++ b/test/fluree/server/integration/basic_transaction_test.clj
@@ -64,6 +64,51 @@
       (is (= {"ledger" ledger-name, "t" 2}
              (-> res :body json/read-value (select-keys ["ledger" "t"])))))))
 
+(deftest ^:integration ^:json transaction-with-dbs-at-multiple-t-values
+  (testing "can transact with a where clause"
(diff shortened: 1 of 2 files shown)


test/fluree/server/integration/basic_transaction_test.clj
```diff
@@ -64,6 +64,51 @@
       (is (= {"ledger" ledger-name, "t" 2}
              (-> res :body json/read-value (select-keys ["ledger" "t"])))))))
 
+(deftest ^:integration ^:json transaction-with-dbs-at-multiple-t-values
```
zonotope, line 67: I think this test could use more English commentary through more nested `testing` macros and/or using the third, `msg` argument of the `is` macro.

zonotope (approved):

zonotope:
Are these changes still relevant in light of https://github.com/fluree/db/pull/1088?

PR #61, opened 2024-05-09
Fix/save raw txn

dpetran:
Fixes https://github.com/fluree/core/issues/105


Jackamus29:
This works great Dan!
The only thing I see is that the txn file has a .json file extension, but in my case with the JWS, it’s really just a text file.

zonotope (approved):

PR #62, opened 2024-05-17
Fix/remove db ledger dependency

bplatz:
This updates to the latest fluree/db sha, and removes a dependency on server that required (:ledger db) to be available which it no longer is.


zonotope (approved):

dpetran (approved):
:100: This looks good.

PR #68, opened 2024-06-05
Handle "fluree-role" header for queries

JaceRockman:
This allows roles to be used in SPARQL queries.


dpetran:
Could you add some tests that demonstrate how this works?

PR #69, opened 2024-06-14
Use new no-consensus by default

bplatz:
This creates a new consensus type of `:none` and updates the default config to use this new type.

Currently raft consensus is run by default even in single-server configurations.

Using no consensus doesn't change much - instead of using raft and multiple communication points for ordering/persisting data redundantly, `:none` consensus uses a simple async chan loop.

Right now it implements all transactions for all local ledgers in a single queue. One obvious improvement would be to spawn separate queues for each ledger as they can be worked on in parallel without issue.

There is definitely some refactoring/improvements that can be made in organizing and reusing consensus code. This does a few things in that direction, but the focus here was a new feature - not a refactor.


zonotope (approved):

src/fluree/server/consensus/none.clj
```diff
@@ -0,0 +1,148 @@
...
+    (let [create-opts    (parse-opts txn)
+          ledger         (deref!
+                          (fluree/create conn ledger-id create-opts))
+          staged-db      (-> ledger
+                             fluree/db
+                             (fluree/stage txn opts)
+                             deref!)
+          commmit-result (deref!
+                         ;; following uses :file-data? and will return map with {:keys [db data-file commit-file]}
+                          (fluree/commit! ledger staged-db {:file-data? true}))]
```
bplatz, line 63: This create+commit process should be replacable with `fluree/create-with-txn`

src/fluree/server/consensus/none.clj
```diff
@@ -0,0 +1,148 @@
...
+                                   (- start-time (:instant params)) "milliseconds.")
+          ledger        (if (deref! (fluree/exists? conn ledger-id))
+                          (deref! (fluree/load conn ledger-id))
+                          (throw (ex-info "Ledger does not exist" {:ledger ledger-id})))
+          staged-db     (-> ledger
+                            fluree/db
+                            (fluree/stage txn opts)
+                            deref!)
+          commit-result (deref!
+                         (fluree/commit! ledger staged-db {:file-data? true}))
```
bplatz, line 93: this stage + commit process can now be replaced with `fluree/transact!`

PR #70, opened 2024-06-24
Use new fluree.db top level api namespace

zonotope:


bplatz (approved):
@zonotope Is it ok to re-target https://github.com/fluree/server/pull/73 to main instead of here and merge? That PR also fixes this cljfmt failing test.

Otherwise we can merge this despite the failing tests, and then merge https://github.com/fluree/server/pull/73.

zonotope:
> @zonotope Is it ok to re-target #73 to main instead of here and merge? That PR also fixes this cljfmt failing test.
> 
> Otherwise we can merge this despite the failing tests, and then merge #73.

I think the best thing is to merge #73 into this branch, wait for ci to pass, and then merge this branch into main to catch any merge issues (not likely, but possible). 

Since #73 is approved, I'll just do that now.

zonotope:
Looks like we have an intermittently failing history test. We can address that separately, so I'm merging now since everything is green.

PR #72, opened 2024-07-16
Refactor/integrant

zonotope:
This patch replaces donut.system with integrant. I believe it simplifies configuration management and the system lifecycle a great deal while avoiding to impose an architecture on the application. 

I decided to make this change because of the experience I had with adding the migrator component the first time and knowing we would have to repeat that process again soon, the difficulty I had with debugging server timeouts and tracing through the execution stack, and because @cap10morgan's good experience with using integrant with components he was working on and my past positive experiences with integrant.


cap10morgan:
Haven't had a chance to review this yet, but I will say that my upcoming nexus component PRs were going to contain a suggestion that we replace donut with integrant everywhere so we aren't dealing with two different approaches / libraries. So I'm very happy to see this!

cap10morgan (approved):
🎊 

(Some formatting and linting issues to address, but looks good once that's done)

PR #74, opened 2024-07-19
Fix policy API tests

bplatz:
With some recent changes, policy related tests were failing.

This is because the tests depend on the `:default-allow? true` setting which was getting set previously.

This updates the failing tests to explicitly include that flag.

The `zero-trust` mode will allow server (and maybe ledger?) level enforcement of this setting, but isn't yet available.


dpetran (approved):
:100:

PR #75, opened 2024-07-20
Standalone consensus component

zonotope:
This patch to the feature/no-consensus branch renames the consensus type from "none" to "standalone" and makes some changes to use explicit configuration values up front instead of passing around config maps, starts and stops the new standalone transactors with integrant, as well as some cleanup.

### DIFF
--- a/resources/config.edn
+++ b/resources/config.edn
@@ -28,9 +28,11 @@
 
  :fluree/subscriptions {}
 
- :fluree/standalone {:conn          #ig/ref :fluree/connection
-                     :watcher       #ig/ref :fluree/watcher
-                     :subscriptions #ig/ref :fluree/subscriptions}
+ :fluree/standalone {:conn            #ig/ref :fluree/connection
+                     :watcher         #ig/ref :fluree/watcher
+                     :subscriptions   #ig/ref :fluree/subscriptions
+                     :max-pending-txs #or [#env FLUREE_STANDALONE_MAX_PENDING_TXS
+                                           16]}
 
  :fluree/handler {:conn          #ig/ref :fluree/connection
                   :consensus     #ig/ref :fluree/consensus
--- a/resources/config.edn
+++ b/resources/config.edn
@@ -32,28 +32,6 @@
                      :watcher       #ig/ref :fluree/watcher
                      :subscriptions #ig/ref :fluree/subscriptions}}
 
- :fluree/raft {:log-history      #or [#env FLUREE_RAFT_LOG_HISTORY
-                                      10]
-               :entries-max      #or [#env FLUREE_RAFT_ENTRIES_MAX
-                                      50]
-               :catch-up-rounds  #or [#env FLUREE_RAFT_CATCH_UP_ROUNDS
-                                      10]
-               :storage-type     #or [#keyword #env FLUREE_RAFT_STORAGE_TYPE
-                                      :file]
-               :servers          #or [#env FLUREE_RAFT_SERVERS
-                                      #profile {:dev    "/ip4/127.0.0.1/tcp/62071"
-                                                :prod   nil
-                                                :docker "/ip4/127.0.0.1/tcp/62071"}]
-               :this-server      #or [#env FLUREE_RAFT_THIS_SERVER
-                                      #profile {:dev    "/ip4/127.0.0.1/tcp/62071"
-                                                :prod   nil
-                                                :docker "/ip4/127.0.0.1/tcp/62071"}]
-               :log-directory    #env FLUREE_RAFT_LOG_DIRECTORY
-               :ledger-directory #env FLUREE_RAFT_LEDGER_DIRECTORY
-               :conn             #ig/ref :fluree/connection
-               :watcher          #ig/ref :fluree/watcher
-               :subscriptions    #ig/ref :fluree/subscriptions}
-
  :fluree/handler {:conn          #ig/ref :fluree/connection
                   :consensus     #ig/ref :fluree/consensus
                   :watcher       #ig/ref :fluree/watcher
--- a/resources/config.edn
+++ b/resources/config.edn
@@ -28,11 +28,7 @@
 
  :fluree/subscriptions {}
 
- :fluree/standalone {:address       #or [#env FLUREE_STANDALONE_ADDRESS
-                                         #profile {:dev    "/ip4/127.0.0.1/tcp/62071"
-                                                   :prod   nil
-                                                   :docker "/ip4/127.0.0.1/tcp/62071"}]
-                     :conn          #ig/ref :fluree/connection
+ :fluree/standalone {:conn          #ig/ref :fluree/connection
                      :watcher       #ig/ref :fluree/watcher
                      :subscriptions #ig/ref :fluree/subscriptions}}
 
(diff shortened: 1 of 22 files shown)


resources/config.edn
```diff
@@ -28,31 +28,9 @@
 
  :fluree/subscriptions {}
 
- :fluree/standalone {:address #or [#env FLUREE_STANDALONE_ADDRESS
-                                   #profile {:dev    "/ip4/127.0.0.1/tcp/62071"
-                                             :prod   nil
-                                             :docker "/ip4/127.0.0.1/tcp/62071"}]}}
- :fluree/raft {:log-history      #or [#env FLUREE_RAFT_LOG_HISTORY
```
bplatz, line 35: I think we want to preserve these settings somewhere, maybe a second config-raft.edn file to at least have them captured.

src/fluree/server/consensus/standalone.clj
```diff
@@ -0,0 +1,135 @@
...
+            [fluree.server.handlers.shared :refer [deref!]]
+            [fluree.server.subscriptions :as subscriptions]
+            [fluree.server.watcher :as watcher]))
+
+(defrecord StandaloneTransactor [tx-queue]
+  consensus/TxGroup
+  (queue-new-ledger [_ ledger-id tx-id txn opts]
+    (go
+      (let [event-msg (msg-format/queue-new-ledger ledger-id tx-id txn opts)]
+        (>! tx-queue event-msg)
```
bplatz, line 19: At some point we need to use `put!` instead of `>!` and maybe you are thinking of that upstream, else we can never return a response and the whole server is locked.

If put! is false, the"requestor" should get a 503 server busy response or equivalent (but the server still might be able to respond to queries, for example). Hopefully it could catch up with backlog and accept new tx requests soon.

If we block all the way up the stack, nothing can happen.

src/fluree/server/consensus/standalone.clj
```diff
@@ -0,0 +1,135 @@
...
+          (cond
+            (= timeout-ch ch)
+            (do
+              (log/trace "No new transactions in 5 second."
+                         "Processed" i "total transactions.")
+              (recur i))
+
+            (nil? event)
+            (do
+              (log/warn "Closing local transaction queue was closed. No new transactions will be processed.")
```
bplatz, line 118: My typo.. but should be:
"Local transaction queue was closed. No new transactions will be processed."

src/fluree/server/consensus/standalone.clj
```diff
@@ -0,0 +1,135 @@
...
+              ::closed)
+
+            :else
+            (let [result (<! (process-event conn subscriptions watcher event))
```
bplatz, line 122: I think we are missing exception handling - which should likely be addressed upstream from this, but this is the point it is swallowed (`process-event` will return `nil`) and things will progress.

The issue is no response will be delivered to the "requestor" that something happened and the will get a timeout response instead.

src/fluree/server/consensus/standalone.clj
```diff
@@ -0,0 +1,135 @@
...
+                                   "with tx-id" tx-id ". Transaction sat in queue for"
+                                   (- start-time (:instant params)) "milliseconds.")
+          ledger        (if (deref! (fluree/exists? conn ledger-id))
+                          (deref! (fluree/load conn ledger-id))
+                          (throw (ex-info "Ledger does not exist" {:ledger ledger-id})))
+          staged-db     (-> ledger
+                            fluree/db
+                            (fluree/stage txn opts)
+                            deref!)
+          commit-result (deref!
```
bplatz, line 83: Per the `process-event` exception handling comment below, here we'll throw and never make it to delivering a message that the HTTP handler is waiting for.

Probably here we don't throw, then deliver the response below (which might be an exception).

Same thing with `create-ledger!` above.

resources/config.edn
zonotope: Yes, either in a separate config raft file, docs, or?

I was planning on starting another branch for config stuff to see how far I get with implementing your suggestions from yesterday.

bplatz:
For your test failure here, not sure but it might need either https://github.com/fluree/db/pull/800 which isn't merged yet. Not certain that is culprit, but think it might be.

src/fluree/server/consensus/standalone.clj
zonotope: I don't think we can use `put!` because we'll run into a "too many pending puts" error under heavy load. 

We could wrap upstream code with a timeout channel to handle returning the 503s, or we could put a dropping buffer on the tx-queue channel and bubble up the return value of `>!` (as long as the return value is false when the buffer is full, but I'd have to verify that this is the case).

bplatz: I must have misunderstood `put!`. I thought it was exactly for this reason and would return false (and not try to `>!` if the buffer was full). I thought there was something like that, maybe a different fn name?

bplatz: Here it is: https://clojuredocs.org/clojure.core.async/offer!

src/fluree/server/consensus/standalone.clj
zonotope: I think I've covered this scenario by [reusing the error handling machinery from the raft implementation](https://github.com/fluree/server/pull/75/files#diff-f93dffc5ef2fc673ab168a1b72e7c8d268aeea3ffb8e92c9760bb23da3384997R75-R78). If there's an error while processing events, then that error is broadcasted out and the assigned watcher is delivered with the error.

resources/config.edn
zonotope: I've saved the raft config options in a separate file. We can merge the config map in that file with the main one, and then `dissoc` the `:fluree/standalone` key and start raft normally. I will wait until we implement the more user friendly configs to make that process more ergonomic.

src/fluree/server/consensus/standalone.clj
```diff
@@ -0,0 +1,113 @@
...
+          (let [error-msg (events/error event-msg result)]
+            (broadcast/announce-error! watcher error-msg))
+          result))
+      (catch Exception e
+        (log/error "Unexpected event message - expected two-tuple of [event-type event-data], "
+                   "and of a supported event type. Received:" event e)))))
+
+(defn new-tx-queue
+  [conn subscriptions watcher]
+  (let [tx-queue (async/chan 16)]
```
zonotope, line 83: I set the buffer size to 16 here to allow for 16 pending transactions at once before we report back that the transactor is too busy. I just made that number up because it kinda felt right, but I have no evidence to back this up. I'm happy to make this number higher or lower. Perhaps we could make it a configuration option that users can set.

bplatz: It is probably more size sensitive than quantity sensitive since it sits in memory, and its a lot of work to estimate/maintain the size so I'm not suggesting we go down that path at least for now.

I don't have a better suggestion, mine would be just a guess too. I'd probably guess a number like 50 though, but have no problem with 16 either.

I think ultimately it should be a server config at the point we have more robust server configs.

bplatz (approved):
LGTM!

src/fluree/server/consensus/standalone.clj
zonotope: It's easy enough to add it as a config option now with a default.

PR #77, opened 2024-07-22
Refactor/dev user ns

bplatz:
This adds some simple REPL startup back to the user namespace, and fixes the http-api calls namespace used for issuing/debugging http calls to the server from the REPL.

I wanted to make sure how this is done makes sense to everyone.

The query-server config has to get re-tested with the changes, and I plan to add a config to startup a query-server only. I don't think anyone has tried that in a while and it doesn't appear the config as it stands is current for that purpose.

I also plan to have a 3-server raft config out of the box startup function similar to what we used to have.

### DIFF
--- a/dev/src/configs.clj
+++ b/dev/src/configs.clj
@@ -1,0 +1,65 @@
+(ns configs
+  (:require [clojure.java.io :as io]
+            [aero.core :as aero]
+            [clojure.string :as str]
+            [meta-merge.core :refer [meta-merge]]
+            [integrant.core :as ig]))
+
+(defn load-config
+  "Loads aero config at given file path using profile keyword."
+  [config-file profile]
+  (-> config-file
+      io/resource
+      (aero/read-config {:profile profile})))
+
+(defn dev-config
+  []
+  (load-config "config.edn" :dev))
+
+;; raft configs
+
+(def server-1 "/ip4/127.0.0.1/tcp/62071")
+(def server-2 "/ip4/127.0.0.1/tcp/62072")
+(def server-3 "/ip4/127.0.0.1/tcp/62073")
+(def servers-str (str/join "," [server-1 server-2 server-3]))
+
+(def raft-single-server
+  (-> (dev-config)
+      (dissoc :fluree/standalone)
+      (assoc :fluree/raft
+             {
+              :storage-type     :file,
+              :ledger-directory "dev/data-raft"
+              :log-directory    "dev/data-raft"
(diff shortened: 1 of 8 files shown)


dev/src/user.clj
```diff
@@ -7,35 +7,46 @@
...
-   :fluree/connection {:storage-path "data/srv3"}
-   :fluree/consensus  {:consensus-servers     servers-str
-                       :consensus-this-server server-3}})
-
-(def query-server-1-overrides
-  {:http/server {:port 58095}
-   :fluree/connection {:method :remote
-                       :servers "http://localhost:58090"}
-   :fluree/consensus {:consensus-type :none}})
+(defn load-config
```
zonotope, line 15: Some of [this functionality will be added in the config branch](https://github.com/fluree/server/compare/feature/user-config?expand=1#diff-2723cd4d3bc8b24ace602f5fbb8ab4e3bd1a67c7142d06bed0c093c777583d47R61-R71) when it's ready. I'm working on validation now, but I hope to be able to open a pr today.

bplatz: Ok, happy to adjust. Just working on REPL setup with the recent changes, and making sure multi-server raft is still working well, which is the more complex REPL setup.

dev/src/configs.clj
```diff
@@ -0,0 +1,65 @@
...
+  (load-config "config.edn" :dev))
+
+;; raft configs
+
+(def server-1 "/ip4/127.0.0.1/tcp/62071")
+(def server-2 "/ip4/127.0.0.1/tcp/62072")
+(def server-3 "/ip4/127.0.0.1/tcp/62073")
+(def servers-str (str/join "," [server-1 server-2 server-3]))
+
+(def raft-single-server
```
zonotope, line 26: I think this stuff could be stored in resource files on the classpath, just like we do the main config. We could add a development resource path under `dev/resources` and call those files `dev-server-1.edn` (and hopefully very soon ".json"), etc and build out functions to merge or override configs in the `fluree.server.system` (and, hopefully very soon, the `fluree.server.config`) namespace. That would make it more robust and also more like what we'll do in production.

dev/src/user.clj
zonotope: I opened #78 to store config in JSON. It might be easier to base this work off of that branch. It's also easier to do merges of configs because we don't have to worry about dissocing irrelevant keys (if you want to replace standalone with raft, for example). though that code isn't there yet, it will be easy to add when we need it.

PR #78, opened 2024-07-24
User JSON Configuration

zonotope:
This patch changes the main configuration file format from edn to json. It also allows users to override default configuration values in two ways: using a "profiles" key in the json configuration, and through environment variables. 

An example configuration file can be found in [resource/config.json](https://github.com/fluree/server/compare/feature/user-config?expand=1#diff-e0011dc26899a9daef5a68c49890208e5a465b163e9394408c11197fc52bf695).

This patch also adds initial validation of user configurations, but the validations are only a first pass and are not as restrictive and precise as they should be, and we should tighten them up later after the config format stabilizes. 

This patch also has some cleanup left over from #75


src/fluree/server/system.clj
```diff
@@ -73,23 +97,16 @@
...
-  ([config]
-   (start-config config {}))
-  ([config overrides]
-   (-> config
-       (meta-merge overrides)
-       ig/init)))
+  [config]
+  (-> config ig/expand ig/init))
 
 (defn start-resource
```
dpetran, line 103: I've got a general question about the config files - if they are `resource`s, how will users who are just working with our jar artifact change them? My understanding (which may be wrong) is that everything in the `resource` dir gets bundled into the jar, and I don't see us looking anywhere else for a config file.

resources/config-raft.json
```diff
@@ -0,0 +1,13 @@
...
+    "logHistory": 10,
+    "entriesMax": 50,
+    "catchUpRounds": 10,
+    "storageType": "file",
+    "servers": "/ip4/127.0.0.1/tcp/62071",
+    "this-server": "/ip4/127.0.0.1/tcp/62071",
+    "log-directory": "data/raft",
+    "ledger-directory": "data/ledger"
+  }
+}
```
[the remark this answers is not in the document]
zonotope: Yes.

src/fluree/server/system.clj
zonotope: That's right. The resource files are the baseline defaults in my mind. You can override those defaults with environment variables only right now, but I think we can add functionality later that allows the user to load a config file from an absolute path on the system pretty easily if we want to do that.

dpetran: I think we may need it sooner than later, @aaj3f was talking about using config files on the docker images, but thanks for confirming my understanding :+1:

zonotope: Ok. I added some command line options parsing and the ability to specify a config file path on the file system to load.

src/fluree/server/config.clj
```diff
@@ -0,0 +1,195 @@
...
+   (m/sequence-schemas)
+   (m/base-schemas)
+   {::path string?
+    ::server-address string?
+    ::connection-storage-method [:enum
+                                 :ipfs :file :memory :s3 :remote]
+    ::indexing-options [:map
+                        [:reindex-min-bytes {:optional true} pos-int?]
+                        [:reindex-max-bytes {:optional true} pos-int?]
+                        [:max-old-indexes {:optional true} pos-int?]]
```
dpetran, line 25: `:max-old-indexes` can be zero, so I think you need `nat-int?` here instead of `pos-int?`

src/fluree/server/main.clj
```diff
@@ -1,14 +1,53 @@
...
+            [fluree.db.util.log :as log]
             [fluree.server.system :as system])
   (:gen-class))
 
 (set! *warn-on-reflection* true)
 
+(def options
+  [["-p" "--profile PROFILE" "Run profile"
+    :parse-fn keyword]
+   ["-c" "--config FILE" "Configuration file path"]
```
dpetran, line 13: Nice, this looks great :+1:

dpetran (approved):
:confetti_ball:

PR #79, opened 2024-07-24
Refactor/dev namespaces

bplatz:
This replaces https://github.com/fluree/server/pull/77 based on the new config feature. It is based on top of https://github.com/fluree/server/pull/78.

This 
- adds some REPL tools to the user namespace
- has a working single-server RAFT config with simple user ns startup
- Adjusts some of our default settings to be more reasonable (open to debate)
- Fixes http-api call dev namespace
- excludes `dev/data` director in .gitignore, and dev namespaces now by default write data to that directory
-

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 !/.clj-kondo/config.edn
 !/.clj-kondo/ci-config.edn
 /.lsp/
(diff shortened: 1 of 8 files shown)


zonotope (approved):
⛵

PR #81, opened 2024-07-29
Better defaults

zonotope:
This small patch changes some defaults for the docker environment. First, it uses 8GB in memory for the main java process, and it logs at the INFO level instead of debug.

We should make some changes in a subsequent pr that makes these entries configurable, but these defaults should suffice for now.


dpetran (approved):
:dango:

PR #82, opened 2024-07-29
Shared LRU cache across connections

bplatz:
This moves the connection setting of `cacheMaxMb` to a server-level setting.

All connections leveraged on the same server should share the same LRU cache.

While technically there is more granular control available, it would in most circumstances under-utilize the machine resources - or if oversized to accommodate that, then leave the server susceptible to OOM failures.

We always have fluree/db, and people can write their own custom thin layer to do anything unique they want.

It utilizes the server-level config options in https://github.com/fluree/server/pull/80

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,7 +1,7 @@
 {:deps {org.clojure/clojure    {:mvn/version "1.11.3"}
         org.clojure/core.async {:mvn/version "1.6.681"}
         com.fluree/db          {:git/url "https://github.com/fluree/db.git"
-                                :git/sha "064d63dd28ea26519c501e1090b8ad18adb58b27"}
+                                :git/sha "55d75e53b2a68edab3d4fd4d96a1c80017d62beb"}
         com.fluree/json-ld     {:git/url "https://github.com/fluree/json-ld.git"
                                 :git/sha "0958995acf5540271d1807fc6d8f2da131164e24"}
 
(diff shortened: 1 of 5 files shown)


zonotope (approved):
This looks good and the rationale makes sense.

One little nitpick is that I think I would rather have individual components for these "server level" configurations instead of one catch-all. That way you could tell from the dependency graph which components depend on what at a more fine-grained level. 

The user level JSON config is fine to have a catch all, but I think it would be nice to have a separate `:fluree/cache` component with an initialization method that created the cache from the `cach-max-mb` argument directly, and all dependent components can `(ig/ref :fluree/cache)` directly.

In any case, this works so I'm fine with it, but I would rather things that need initialization live in their own integrant components, even if they are grouped under a single key in the user-level json config.

bplatz:
> In any case, this works so I'm fine with it, but I would rather things that need initialization live in their own integrant components, even if they are grouped under a single key in the user-level json config.

@zonotope I believe this addresses your comment: https://github.com/fluree/server/pull/82/commits/fcf570c4518ea78f421c98e8a07b834409077781

zonotope:
> > In any case, this works so I'm fine with it, but I would rather things that need initialization live in their own integrant components, even if they are grouped under a single key in the user-level json config.
> 
> @zonotope I believe this addresses your comment: [fcf570c](https://github.com/fluree/server/commit/fcf570c4518ea78f421c98e8a07b834409077781)

Looks good to me.

PR #83, opened 2024-08-02
Fix/profile argument

zonotope:
This small patch updates the docker "CMD" string to use the updated fluree/server argument for specifying the profile. It also cleans up the passed in profile string to trim any surrounding whitespace.


dpetran (approved):
:scroll:

PR #88, opened 2024-08-22
Feature/import api

bplatz:
This adds a new 'import' API to fluree/server which accepts TURTLE (.ttl) and JSON-LD.

The main feature of this is that raw RDF data can be imported without converting it to JSON-LD... and even raw JSON-LD can be imported without having to reformat it by wrapping your JSON-LD file in {"insert": <json-ld>}.

The RDF data is just sent in the HTTP body, regardless of format.

To use the new API with TURTLE, the HTTP Headers must include:
```json
{"Content-Type":  "test/turtle",
 "Fluree-Ledger": "<fluree-ledger-name-here>"}
```

To use the API with JSON-LD, the HTTP Headers must include:
```json
{"Content-Type":  "application/json",
 "Fluree-Ledger": "<fluree-ledger-name-here>"}
```

For now, the ledger must already exist, and you are just 'importing' into an existing ledger. Having the ability to create a new ledger from an import is likely desirable, and it can be added fairly simply.

### DIFF
--- a/src/fluree/server/handler.clj
+++ b/src/fluree/server/handler.clj
@@ -190,6 +190,13 @@
         (handler req*))
       (handler (assoc req :raw-txn body-params)))))
 
+(defn wrap-set-ledger-id
+  [handler]
+  (fn [req]
+    (->> (get-in req [:headers "fluree-ledger"])
+         (assoc req :fluree/ledger-id)
+         (handler))))
+
 (defn wrap-set-fuel-header
   [handler]
   (fn [req]
@@ -334,6 +341,7 @@
                                    [10 wrap-cors]
                                    [10 (partial wrap-assoc-system conn consensus watcher)]
                                    [50 unwrap-credential]
+                                   [75 wrap-set-ledger-id]
                                    [100 wrap-set-fuel-header]
                                    [200 coercion/coerce-exceptions-middleware]
                                    [300 coercion/coerce-response-middleware]
--- a/src/fluree/server/handler.clj
+++ b/src/fluree/server/handler.clj
@@ -165,26 +165,30 @@
   as :body-params. If the signature is not valid, throws an invalid signature error. If
   the request body is not a credential, nothing is done."
   [handler]
-  (fn [{:keys [body-params] :as req}]
+  (fn [{:keys [body-params content-type] :as req}]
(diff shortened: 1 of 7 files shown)


src/fluree/server/handler.clj
```diff
@@ -165,26 +171,37 @@
...
-            {:subject body-params}
-
-            :else ; invalid credential
-            (throw (ex-info "Invalid credential"
-                            {:response {:status 400
-                                        :body   {:error "Invalid credential"}}})))
-          req*     (assoc req :body-params subject :credential/did did :raw-txn body-params)]
-      (log/debug "Unwrapped credential with did:" did)
-      (handler req*))))
+    (if (#{"application/json" "application/jwt"} content-type)
```
dpetran, line 176: Is it possible to sign a turtle import? I think you'd use a JWT, but `cred/verify` will not handle turtle correctly.

src/fluree/server/handler.clj
```diff
@@ -62,6 +62,12 @@
 (def TransactRequestBody
   (m/schema [:map-of :any :any]))
 
+(def ImportRequestBody
+  (m/schema [:or
+             [:string {:min 1}]
+             [:map-of :any :any]
+             [:sequential map?]]))
```
dpetran, line 69: Just checking to make sure I understand, these schemas correspond to:
- turtle string
- json-ld map
- vector of json-ld maps

Is that right?

dpetran (commented):
I wonder if we should name this `add` instead of `import`, sort of mirroring the [SPARQL Update API](https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#add). It leaves room for us to follow those conventions for other graph-level operations. We already have `create`, we could add `copy`, `move`, and `drop` as well, operations that I think could be especially useful for doing git-like branching workflows, or moving graphs between different types of storage.

src/fluree/server/handler.clj
bplatz: I think so, but it wasn't the priority and I think there is a good chance this middleware dissapears, but not 100% on that.

The db api now has `*-credential` APIs which ideally we'd mostly leverage instead of fluree/server doing its own thing.

But for now it doesn't have signatures for TTL imports, although I should make a ticket for it as I think it should at some point.

PR #89, opened 2024-09-05
add extra error logging and take from migration channel

dpetran:
Add extra logging for failures and add a take from the migration channel.

### DIFF
--- a/src/fluree/server/task/migrate_sid.clj
+++ b/src/fluree/server/task/migrate_sid.clj
@@ -1,6 +1,7 @@
 (ns fluree.server.task.migrate-sid
   (:require [clojure.core.async :as async :refer [<! >! go-loop]]
             [fluree.db.json-ld.migrate.sid :as migrate-sid]
+            [fluree.db.util.async :refer [go-try]]
             [fluree.db.util.core :as util]
             [fluree.db.util.log :as log]))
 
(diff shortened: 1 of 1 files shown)


src/fluree/server/task/migrate_sid.clj
```diff
@@ -19,20 +18,27 @@
...
-            error-ch       (async/chan)]
-        (log-migrated-index-files index-files-ch error-ch)
-        (if alias
-          (let [_ (log/info :migrate/sid "ledger" alias "start" (util/current-time-iso))
-                migrate-ch (migrate-sid/migrate conn alias nil force index-files-ch)]
-            (async/alt!
-              error-ch ([e]
-                        (throw e))
+  (log/info :migrate/sid "start" (util/current-time-iso) "." ledgers force)
+  (try
```
bplatz, line 22: This `try` will never catch anything. The return value of `async/<!!` could be an exception, but won't be a thrown exception.

You'll need to check the return value, and if an exception then `throw` it for this `try` to actually catch it (which is what the `go-try` macro does). But at that point you don't need the `try`.

If the main purpose of the `throw` statements below is to exit the loop, I think you simply make `util/exception?` checks you do below not call recur.

Not sure if there is anything to catch an exception upstream, but if so you'll need to either do what I describe above or use `go-try`.

dpetran: Thanks for catching that, I was trying to get this up for Jake quickly last week and I changed it from a `go-try` to a `go-loop` while doing build debugging. I've put it back to the way it was.

bplatz (approved):
LGTM!

PR #91, opened 2024-09-12
update json-ld and db deps - prevent variable expansion

dpetran:


deps.edn
```diff
@@ -1,9 +1,9 @@
 {:deps {org.clojure/clojure    {:mvn/version "1.11.3"}
         org.clojure/core.async {:mvn/version "1.6.681"}
         com.fluree/db          {:git/url "https://github.com/fluree/db.git"
-                                :git/sha "3c403fd79164a09697c0a0e9bb7ff487959c0fcb"}
+                                :git/sha "1eb181cb3c66f865c77493765d8d429af4359755"}
         com.fluree/json-ld     {:git/url "https://github.com/fluree/json-ld.git"
-                                :git/sha "0958995acf5540271d1807fc6d8f2da131164e24"}
+                                :git/sha "7ae310b2d8c0aaf10e107e094607f69464ed3447"}
```
bplatz, line 6: Wouldn't we be better off not having this as a dependency here at all, and only in fluree/db?

I believe with both, we don't have control over which one is used if they are different.

bplatz (approved):
:pizza:

deps.edn
dpetran: I thought about that as I bumped it, but if we're using it directly I think we should depend on it directly. The build process does the right thing with regard to transitive deps, so there's no size/perf downside to specifying it either.

PR #93, opened 2024-09-17
Feature/closed mode

dpetran:
https://github.com/fluree/core/issues/112

This adds closed-mode support to server.

There are now two new config keys:
- ["http", "rootIdentities"]: an array of keypair ids
- ["http", "closedMode"]: a boolean flag to enable `closed-mode`

When closed mode is enabled:
- every request is required to be signed
- only a root identity is allowed to create ledgers
- non-root identities are not allowed to claim other identities

In this implementation it is specified that requests from root identities will NOT be called with `wrap-policy`. However, `server` does not call `wrap-policy` on its own, so I've left that for future work.

### DIFF
--- a/resources/config.json
+++ b/resources/config.json
@@ -22,7 +22,7 @@
     "server": "jetty",
     "port": 8090,
     "maxTxnWaitMs": 120000,
-    "rootIdentities": [],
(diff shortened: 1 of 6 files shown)


zonotope (approved):
🚪

PR #96, opened 2024-10-04
Federated queries across servers

zonotope:
This patch integrates fluree/db#906 to implement federated queries across distinct server instances. Please see [fluree.server.test.integration.remote-system-test](https://github.com/fluree/server/blob/ec964f6040d7397879582f5e45ea50fbfbe9d42a/test/fluree/server/integration/remote_system_test.clj) for a demonstration of three separate server instances running on 3 different ports that are each able to pull data in from the others when processing queries. 

We needed to make fluree connections and server instances much more flexible to facilitate this functionality, so as part of this work I've also added a new json-ld based configuration format that allows users to describe different configurations and how they fit together using the same format that the data in the database is stored in.


zonotope:
@fluree/core There are 6 test failures here, but they are unrelated to this work. Updating the db dependency to the latest on fluree/server main HEAD has the same failures. I will try to get to the bottom of those failures later.

src/fluree/server/system.clj
```diff
@@ -13,129 +27,619 @@
...
-(defmethod ig/init-key :fluree/sid-migration
+(defmethod ig/init-key :fluree.server/sid-migration
   [_ {:keys [conn ledgers force]}]
   (task.migrate-sid/migrate conn ledgers force))
 
+(defmethod ig/init-key :default
+  [_ config]
+  config)
+
+(def default-resource-name "config.jsonld")
```
aaj3f, line 625: @zonotope Right now it seems `config.jsonld` is an example of what's possible (e.g. with the s3 & ipns publishers). But since `resources/config.jsonld` is the default that's loaded if no config is passed, we may want to create a new file like `resources/full-example-config.jsonld` and then make `config.jsonld` a more representative default. Maybe something like this

```json
{
  "@context": {
    "@base": "https://ns.flur.ee/dev/config/main/",
    "@vocab": "https://ns.flur.ee/system#",
    "profiles": {
      "@container": [
        "@graph",
        "@id"
      ]
    }
  },
  "@id": "standaloneServer",
  "@graph": [
    {
      "@id": "localDiskStorage",
      "@type": "Storage",
      "addressIdentifier": "ee.flur/main",
      "filePath": "/opt/fluree-server/data"
    },
    {
      "@id": "connection",
      "@type": "Connection",
      "parallelism": 4,
      "cacheMaxMb": 200,
      "commitStorage": {
        "@id": "localDiskStorage"
      },
      "indexStorage": {
        "@id": "localDiskStorage"
      },
      "primaryPublisher": {
        "@type": "Publisher",
        "storage": {
          "@id": "localDiskStorage"
        }
      }
    },
    {
      "@id": "consensus",
      "@type": "Consensus",
      "consensusProtocol": "standalone",
      "maxPendingTxns": 42,
      "connection": {
        "@id": "connection"
      }
    },
    {
      "@id": "http",
      "@type": "API",
      "httpPort": 8090,
      "maxTxnWaitMs": 120000
    }
  ]
}
```

Currently if you run `server` as an uberjar with no config option or env vars passed in, it fails w/ aws client connect errors, I'm assuming because the default `config.jsonld` includes those hypothetical examples

zonotope: > @zonotope Right now it seems `config.jsonld` is an example of what's possible (e.g. with the s3 & ipns publishers). But since `resources/config.jsonld` is the default that's loaded if no config is passed, we may want to create a new file like `resources/full-example-config.jsonld` and then make `config.jsonld` a more representative default. Maybe something like this
> 
> ```json
> {
>   "@context": {
>     "@base": "https://ns.flur.ee/dev/config/main/",
>     "@vocab": "https://ns.flur.ee/system#",
>     "profiles": {
>       "@container": [
>         "@graph",
>         "@id"
>       ]
>     }
>   },
>   "@id": "standaloneServer",
>   "@graph": [
>     {
>       "@id": "localDiskStorage",
>       "@type": "Storage",
>       "addressIdentifier": "ee.flur/main",
>       "filePath": "/opt/fluree-server/data"
>     },
>     {
>       "@id": "connection",
>       "@type": "Connection",
>       "parallelism": 4,
>       "cacheMaxMb": 200,
>       "commitStorage": {
>         "@id": "localDiskStorage"
>       },
>       "indexStorage": {
>         "@id": "localDiskStorage"
>       },
>       "primaryPublisher": {
>         "@type": "Publisher",
>         "storage": {
>           "@id": "localDiskStorage"
>         }
>       }
>     },
>     {
>       "@id": "consensus",
>       "@type": "Consensus",
>       "consensusProtocol": "standalone",
>       "maxPendingTxns": 42,
>       "connection": {
>         "@id": "connection"
>       }
>     },
>     {
>       "@id": "http",
>       "@type": "API",
>       "httpPort": 8090,
>       "maxTxnWaitMs": 120000
>     }
>   ]
> }
> ```
> 
> Currently if you run `server` as an uberjar with no config option or env vars passed in, it fails w/ aws client connect errors, I'm assuming because the default `config.jsonld` includes those hypothetical examples

Yes, we should definitely do this. I was planning to in the final cleanup push, but I hadn't had a chance to yet.

zonotope: We should also convert the raft example too.

resources/config.jsonld
```diff
@@ -0,0 +1,50 @@
+{
+  "@context": {
```
dpetran, line 2: Do we still support a regular json config file, or do we require users to update their config?

dpetran (commented):
I'm curious about why we've moved the config to be json-ld. It seems like a source of significant additional complexity.

resources/config.jsonld
zonotope: user's will need to update their config.

zonotope:
> I'm curious about why we've moved the config to be json-ld. It seems like a source of significant additional complexity.

The additional complexity in the configuration format was necessitated by the additional flexibility in the systems the configuration format needed to support. Any configuration format that wasn't JSON-LD and yet was flexible enough to support starting all the different types of systems we need to start would just be an ad-hoc re-implementation of JSON-LD and yet another syntax for our users to learn. This is a JSON-LD database, so that's the one configuration format we know our users either already know or soon will. 

We can have pre-built configs for common scenarios, but we needed a richer format for the general case, and all the essential required features just happened to be supported by JSON-LD.

dpetran (approved):
:100:

PR #98, opened 2024-10-11
Add policy Fluree-Policy* headers for policy SPARQL

bplatz:
Adds support for 4 new HTTP headers to control policy.

These are here primarily for SPARQL queries which don't have an `opts` map where we can define policy-related items.

The headers added, along with their equivalent `:opts` map keys are:
`Fluree-Policy-Class` -> `:policy-class`
`Fluree-Policy-Identity` -> `:identity`
`Fluree-Policy-Values` -> `:policy-values` (stringified JSON)
`Fluree-Policy` -> `:policy` (stringified JSON)

*Note these will be overridden by a signed credential request.

Tests are included for all of these headers.

### DIFF
--- a/src/fluree/server/handler.clj
+++ b/src/fluree/server/handler.clj
@@ -260,10 +260,10 @@
                                                 {:status 400})))))]
       (handler
        (cond-> req
-               policy-identity (assoc :policy/identity policy-identity)
-               policy-class (assoc :policy/class policy-class)
-               policy (assoc :policy/policy policy)
-               policy-values (assoc :policy/values policy-values))))))
+         policy-identity (assoc :policy/identity policy-identity)
+         policy-class (assoc :policy/class policy-class)
+         policy (assoc :policy/policy policy)
+         policy-values (assoc :policy/values policy-values))))))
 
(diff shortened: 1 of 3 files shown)


src/fluree/server/handler.clj
```diff
@@ -231,6 +232,39 @@
...
+  that might have otherwise been set in FlureeQL/JSON."
+  [handler]
+  (fn [req]
+    (let [policy-identity (get-in req [:headers "fluree-policy-identity"])
+          policy-class    (get-in req [:headers "fluree-policy-class"])
+          policy          (when-let [p (get-in req [:headers "fluree-policy"])]
+                            (try
+                              (json/parse p false)
+                              (catch Exception _
+                                (throw (ex-info "Invalid policy header, it must be parsable JSON."
```
dpetran, line 249: A more precise error message might be more useful: "Invalid fluree-policy-class header: must be JSON".

dpetran (approved):
:sparkle:

src/fluree/server/handler.clj
bplatz: Done!

PR #101, opened 2024-11-28
Allow environment variables in server configs

zonotope:
This patch removes the majority of the configuration processing and parsing code in favor of that added to the db library in fluree/db#943. The only configuration parsing and initialization code left here is that which is specific to fluree/server. This patch also makes use of the environment variable aware configuration parsing code added in fluree/db#943 so that users can specify that certain values should be loaded from environment variables.


dpetran (approved):
:100:

kurtharriger:
Looks like user ns still references recently removed config methods causing the repl not to start for me 
https://github.com/fluree/server/blob/feature/config-profiles/dev/src/user.clj#L14

zonotope:
> Looks like user ns still references recently removed config methods causing the repl not to start for me https://github.com/fluree/server/blob/feature/config-profiles/dev/src/user.clj#L14

I've fixed this now.

PR #107, opened 2025-01-15
Fix Hanging Transactions

zonotope:
This patch pulls in @aaj3f's excellent automated benchmarking namespace and updates the fluree/db dependency to bring in the query optimization merged in fluree/db#950. This speeds up both queries and transactions dramatically.


dpetran (approved):
:chair:

PR #108, opened 2025-01-20
add support for empty result set with selectOne

dpetran:
I've mapped a `nil` result when a `selectOne` query returns no results to the JSON string "null", which parses to `null` in Javascript. Without this support the response fails to conform to the `QueryResponse` schema.

### DIFF
--- a/src/fluree/server/handler.clj
+++ b/src/fluree/server/handler.clj
@@ -93,7 +93,7 @@
 (def QueryResponse
   (m/schema [:orn
              [:select [:sequential [:or coll? map?]]]
-             [:select-one [:or coll? map?]]]))
+             [:select-one [:or coll? map? nil? [:enum "null"]]]]))
 
 (def HistoryQuery
   (m/schema (fqh/history-query-schema [[:from LedgerAlias]])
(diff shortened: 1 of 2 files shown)


src/fluree/server/handler.clj
```diff
@@ -248,6 +248,15 @@
                   (handler req*))))
         (handler req)))))
 
+(def wrap-no-result
+  "Handle a nil response from a selectOne query with no results."
+  (fn [handler]
+    (fn [req]
+      (let [{:keys [status body] :as result} (handler req)]
+        (if (and (= 200 status) (nil? body))
```
zonotope, line 256: This feels like something that should be handled by muuntaja instead of rolling our own conversion for this specific case. Is there no way to accomplish this using that library since we're already using it to convert between formats? I can see this causing issues if we ever decide to respond with edn as well as json.

dpetran: I gave it an honest try with muuntaja but wasn't willing to devote more than 2 hours to it. Once I couldn't get it working in that window I went back to this. I figure that if we ever do support edn there will need to be a lot of updates to the middleware anyways.

zonotope: Is this a problem with the `QureryResponse` schema? What happens when you just account for the possibility that `:select-one` query responses can also be `nil`?

dpetran: Then it converts the `:body` key to the empty string `""`, which isn't valid json.

PR #109, opened 2025-01-22
Fix existing ledger instantiation

zonotope:
This patch updates the fluree/db dependency sha to pick up the fix in fluree/db#955 and adds a few useful namespace requires to the fluree server repl.


dpetran (approved):
:green_book:

PR #113, opened 2025-02-13
Fix remote nameservices

zonotope:
This patch updates the db dependency to bring in the fix from fluree/db#968


dpetran (approved):
:mountain:

PR #115, opened 2025-02-20
Broadcast Protocol

zonotope:
This patch has changes that make it easier to use Server for the Nexus use case. 
- adds a `Broadcaster` protocol to better abstract reporting transactions to interested clients
- adds iris for broadcaster properties and types to the configuration vocabulary
- refactors the command line options parsing so it can be used by dependent systems
- simplifies the consensus event message format and makes it more consistent by using a single map with a type key for every message instead of inconsistently using two element vectors with a type as the first entry an extra data map as the second

It also includes some cleanup, reorganization, and improvements:
- removes the redundancy from the server entry point namespace by changing it to `fluree.server` from `fluree.server.main`.
- moves broadcast related namespaces out of the consensus hierarchy
- changes the `Makefile` to be self-documenting, and adds a `help` task to list and describe the available tasks, like this:
```
$ make help
benchmark       Benchmark performance
clean           Remove build artifacts
cljfmt-check    Check Clojure formatting with cljfmt
clj-kondo-lint  Lint Clojure code with clj-kondo
docker-build    Build the server docker container
docker-push     Build and publish the server docker container
docker-run      Run in docker
help            Describe available tasks
pending-tests   Run pending tests
prepare         Prepare clojure dependencies
test            Run tests
uberjar         Build an executable server uberjar
```


dpetran (approved):
:tokyo_tower:

PR #116, opened 2025-02-21
Unblock transaction queue in the case of errors

zonotope:
Previously an error while processing an event message sent to the transaction queue would attempt to put a `nil` value on the transaction queue's output channel, which generated an uncaught exception that would effectively block the transaction queue from processing further messages. 

This patch detects error/`nil` values while processing events and prevents those values from going to the output channel so that subsequent events can still be processed.


dpetran (approved):
:nail_care:

PR #117, opened 2025-02-28
Feature/sparql protocol query

dpetran:
This adds support for `Accept: application/sparql-results+json` headers, making our `fluree/query` endpoint a SPARQL Protocol compliant endpoint. 

I also added support for all query opts being specified as http headers in order to bring feature parity to SPARQL queries. Options specified in headers take precedence over those specified in the query body for FQL.

Related to https://github.com/fluree/core/issues/170

### DIFF
--- a/src/fluree/server/handler.clj
+++ b/src/fluree/server/handler.clj
@@ -251,17 +251,20 @@
           fuel 1000] ; TODO: get this for real
       (assoc-in resp [:headers "x-fdb-fuel"] (str fuel)))))
 
+(def fluree-header-opts
+  ["fluree-meta" "fluree-max-fuel" "fluree-identity" "fluree-policy-identity"
+   "fluree-policy" "fluree-policy-class" "fluree-policy-values"
+   "fluree-format" "fluree-output"])
+
 (defn wrap-header-opts
   "Extract options from headers, parse and validate them where necessary, and attach to
   request. Opts set in the header override those specified within the transaction or
-  query. Intended mainly to provide a method of specifying opts for SPARQL queries"
+  query."
   [handler]
   (fn [{:keys [headers credential/did] :as req}]
     (let [{:keys [meta max-fuel identity policy-identity policy policy-class policy-values format output]}
           (-> headers
-              (select-keys ["fluree-meta" "fluree-max-fuel" "fluree-identity" "fluree-policy-identity"
-                            "fluree-policy" "fluree-policy-class" "fluree-policy-values"
-                            "fluree-format" "fluree-output"])
+              (select-keys fluree-header-opts)
               (update-keys (fn [k] (keyword (subs k (count "fluree-"))))))
 
           meta     (when meta
--- a/src/fluree/server/handler.clj
+++ b/src/fluree/server/handler.clj
@@ -1,5 +1,6 @@
 (ns fluree.server.handler
   (:require [clojure.core.async :as async :refer [<!!]]
+            [clojure.string :as str]
             [fluree.db.json-ld.credential :as cred]
             [fluree.db.query.fql.syntax :as fql]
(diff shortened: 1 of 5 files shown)


dpetran:
fixes https://github.com/fluree/core/issues/171

dpetran:
Rebased on main.

src/fluree/server/handler.clj
```diff
@@ -255,38 +256,74 @@
...
-(defn wrap-policy-metadata
-  "Both Fluree-Policy-Class and Fluree-Policy-Identity can be optionally
-  set in the header for non-credentialed queries/transactions. The primary
-  motivation is enforcing policy with SPARQL which doesn't have an opts
-  map option, although if set, it will also override policyClass/Identity
-  that might have otherwise been set in FlureeQL/JSON."
+(defn wrap-header-opts
+  "Extract options from headers, parse and validate them where necessary, and attach to
+  request. Opts set in the header override those specified within the transaction or
+  query. Intended mainly to provide a method of specifying opts for SPARQL queries"
```
zonotope, line 262: Super minor, but I would drop this last sentence since these opts apply generally.

src/fluree/server/handler.clj
```diff
@@ -255,38 +256,74 @@
...
-      (handler
-       (cond-> req
-         policy-identity (assoc :policy/identity policy-identity)
-         policy-class (assoc :policy/class policy-class)
-         policy (assoc :policy/policy policy)
-         policy-values (assoc :policy/values policy-values))))))
+  (fn [{:keys [headers credential/did] :as req}]
+    (let [{:keys [meta max-fuel identity policy-identity policy policy-class policy-values format output]}
+          (-> headers
+              (select-keys ["fluree-meta" "fluree-max-fuel" "fluree-identity" "fluree-policy-identity"
```
zonotope, line 267: I think these recognized keys should be defined in a separate top level binding so we can easily reference the recognized header options.

src/fluree/server/handler.clj
```diff
@@ -255,38 +256,74 @@
...
+                            "fluree-policy" "fluree-policy-class" "fluree-policy-values"
+                            "fluree-format" "fluree-output"])
+              (update-keys (fn [k] (keyword (subs k (count "fluree-"))))))
+
+          meta     (when meta
+                     (case meta
+                       "false" false
```
zonotope, line 274: I think it's friendlier to run `clojure.string/lower-case` on this value first.

src/fluree/server/handler.clj
```diff
@@ -255,38 +256,74 @@
...
+                              (= output "fql")    :fql
+                              :else               :fql)
+          policy        (when policy
+                          (try (json/parse policy false)
+                               (catch Exception _
+                                 (throw (ex-info "Invalid Fluree-Policy header: must be JSON."
+                                                 {:status 400})))))
+          policy-values (when policy-values
+                          (try (let [pv (json/parse policy-values false)]
+                                 (if (sequential? pv)
```
zonotope, line 304: Does it make sense to validate this with malli since we already have a validator for that defined?

zonotope (approved):
I have a few suggestions, but looks good.

src/fluree/server/handler.clj
dpetran: I think the original idea was that the whole query will be run through the coercer/validator down the line, so we just do a very cheap check right here. If it's malformed we'll get a proper error message when the whole query is checked.

dpetran:
This PR depends on a couple of companion PRs in `db`, so I'm going to wait on merging this until those have been reviewed:
https://github.com/fluree/db/pull/973
https://github.com/fluree/db/pull/985

dpetran:
Rebased on main.

PR #118, opened 2025-02-28
add open telemetry and structured json logging

kurtharriger:
This adds the Open Telemetry (Otel) agent and structured logging.  

Otel agent will automatically instruments several java libraries such as jetty with various metrics such as request duration and distributed tracing for xray.  However, otel uses thread local storage and thus any thread pools need to be wrapped to preserve trace context and I've monkey patched core async thread pools as necessary to propagate the state.  When the trace context is available it should in most cases be automatically be propagated to downstream services 
See https://github.com/steffan-westcott/clj-otel/blob/master/doc/guides.adoc#manually-propagate-context-in-an-http-client-request

However there are still some cases such as internal message queues such as consensus where the trace context needs to be manually propagated.  I've updated the standalone consensus algorithm to start but there are likely other places like raft and broadcaster where it may make sense to propagate the state also.  It may also be useful to add additional trace annotations for context.  I've added ledger and transaction id in the standalone trace but we could consider adding more of these attributes at the start of the request.  See https://github.com/steffan-westcott/clj-otel/blob/master/doc/guides.adoc#add-attributes-to-a-span

This also enables reporting of additional arbitrary metrics to and build dashboards such as total fuel consumed or whatever we want to publish. See  https://github.com/steffan-westcott/clj-otel/blob/master/doc/guides.adoc#add-manual-metrics-instrumentation

Metrics and tracing will require some additional deployment configuration after this is merged before they show up in aws, but the docker image has been updated to use structured logging on the next deployment and will include the instrumentation so that trace ids also show up in the logs.  

For production its important to have each log on its own line and preferably json which is now enabled, but for local development readability is better so I still use the plain text rather than structured logging for readability, but I've tweaked the default logging pattern a bit such that the metadata and full logger is placed on one line and the message below on the next line. I don't like the abbreviated logger names and I find it much easier to read with metadata on one line and log message on the next.  I removed the dev/logback.xml and used some env vars to enable dynamic configuration of appender and log levels such that this seemed unnecessary.   

To validate metrics and traces appear as desired locally I've also added a docker compose file in extras/otel that will run grafana, prometheus, jaeger and loki.  When you run clojure -X:run-dev no instrumentation is added and thus trace ids will not appear in the logs.  If you would like to see trace ids in the logs you can run the `docker-compose up` in the extras/otel folder and then `extras/otel/start-dev.sh` or `extras/otel/start-dev.sh  repl` if you want to start the repl and attach to it.


kurtharriger:
I'm going to move this to nexus and validate there before we OSS

PR #119, opened 2025-03-01
Latest fluree/db, no need for 'prepare'

bplatz:
This removes the 'prepare' steps in the makefile, as with the latest db it is no longer needed.

Updates to latest fluree/db:main

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /usr/src/fluree-server
 
 COPY deps.edn ./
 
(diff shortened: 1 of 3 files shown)


kurtharriger (commented):
There is also a prep in the Dockerfile

https://github.com/fluree/server/blob/03e2936110c9278fd5a2d728174cacff22e06e7e/Dockerfile#L8

zonotope (approved):
🏉

PR #120, opened 2025-03-01
Updates to work with no fluree ns needed db

bplatz:
This updates fluree/server to work with https://github.com/fluree/db/pull/983


zonotope (approved):
🏹

PR #121, opened 2025-03-11
Separate broadcast and watcher

zonotope:
This patch separates the transaction watching logic, which is mainly to allow processes local to the server instance to monitor a transaction, from the broadcasting logic, which allows connected systems to monitor the changes to a ledger. Separating those two abstractions allows them to be used and built upon more easily independently. 

This patch also provides a method to inject custom routes into the request handler at start time to allow for customizing the types of requests a server deployment can accept. 

Lastly, this patch refactors some of the command processing code for easier reuse.


dpetran (approved):
:satellite:

PR #126, opened 2025-03-21
add support for application/sparql-update transaction requests

dpetran:
Implements the SPARQL Protocol [update operation](https://www.w3.org/TR/2013/REC-sparql11-protocol-20130321/#update-operation).

In lieu of pushing down sparql translation to the `db` api, this PR just translates the SPARQL update into fql at the first opportunity in order to minimize structural changes to transaction processing which depends on txn introspection.

Depends on https://github.com/fluree/db/pull/992

https://github.com/fluree/core/issues/175

### DIFF
--- a/src/fluree/server/handlers/transact.clj
+++ b/src/fluree/server/handlers/transact.clj
@@ -105,7 +105,7 @@
 (defhandler default
   [{:keys [fluree/conn fluree/consensus fluree/watcher fluree/broadcaster credential/did fluree/opts raw-txn]
     {:keys [body]} :parameters}]
-  (let [txn            (if (= :sparql (:format opts))
+  (let [txn            (if (sparql/sparql-format? opts)
                          (sparql/->fql body)
                          body)
         txn-context    (ctx-util/txn-context txn)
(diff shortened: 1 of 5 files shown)


src/fluree/server/handlers/transact.clj
```diff
@@ -102,13 +103,16 @@
                                 :body   {:error err-message}}}))))
 
 (defhandler default
-  [{:keys [fluree/conn fluree/consensus fluree/watcher fluree/broadcaster credential/did raw-txn]
+  [{:keys [fluree/conn fluree/consensus fluree/watcher fluree/broadcaster credential/did fluree/opts raw-txn]
     {:keys [body]} :parameters}]
-  (let [txn-context    (ctx-util/txn-context body)
-        ledger-id      (transact-api/extract-ledger-id body)]
+  (let [txn            (if (= :sparql (:format opts))
```
zonotope, line 108: I think you should use the function you just defined in db here.

Also, maybe not for this pr, but I think we should eventually write a function that can extract the ledger id directly from a sparql query, or move all sparql conversion outside of the db library here. (This is just a gut level thought. I haven't thought through all of the implications yet)

zonotope (approved):
🍃

PR #128, opened 2025-03-26
Feature/consensus broadcast

zonotope:
This patch changes the responsibility of broadcasting a new commit (back) to the transactor instead of the requesting server/process. There was the same amount of complexity to communicate the new transaction back to the requesting process as there would have been to broadcast the message in the first place, plus the extra complexity to broadcast the message out from the requesting process.

This patch also simplifies the broadcast and watcher apis to use generic events, and has some cleanup and cljfmt fixes.


dpetran (approved):
:telephone_receiver:

PR #130, opened 2025-04-09
ensure shacl shapes are used for validation on load

dpetran:
Properly handle loading a ledger with SHACL shape validation. The actual change is in https://github.com/fluree/db/pull/995


zonotope (approved):
🏟️

PR #131, opened 2025-04-14
Feature/store transaction

zonotope:
This patch adds functions to transform consensus events to store any transactions they contain to connection storage and replace those transactions with transaction addresses. It also changes the standalone transactor to check if any transactions need to be resolved from storage (and resolve them if necessary) before attempting to process consensus events. It also simplifies the transaction/create handlers and the transaction processing to rely more on the default behavior of the db library instead of checking whether ledgers exist, and stage and committing transactions instead of transacting directly here when that's already done by the db library.


src/fluree/server/consensus/standalone.clj
```diff
@@ -40,43 +30,47 @@
...
-          ledger        (if (deref! (fluree/exists? conn ledger-id))
-                          (deref! (fluree/load conn ledger-id))
-                          (throw (ex-info "Ledger does not exist" {:ledger ledger-id})))
-          staged-db     (-> ledger
-                            fluree/db
-                            (fluree/stage txn opts)
-                            deref!)
-          commit-result (deref! (fluree/apply-stage! ledger staged-db))]
+          opts*         (update opts :meta ensure-file-reports)
+          commit-result (deref! (fluree/transact! conn txn opts*))]
```
dpetran, line 34: Nice to see we're finally using this api in server!

dpetran (approved):
:convenience_store:

PR #134, opened 2025-04-18
Publish Release on Tag

kurtharriger:
When tag is pushed it will push docker image to docker hub and create a github release with change log and link to the docker image.  See https://github.com/fluree/server/releases

If the tag is on the main branch it should also update the fluree/server:latest tag if version number does not have a suffix so that it does not get updated for prerelease tags like v1.2-alpha or v1.2-experimentalfeature.
Also allows creating branch of older version such as release/v1.1 for support and maintenance of older releases if needed without erroneously updating latest (since these hotfixes may be backported from main but not on main) 

After review and merge I'll tag as v1.0.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,81 +1,154 @@
 name: CI
 on:
-  push:
-  schedule:
-    - cron: 13 21 * * *
+ push:
+  branches:
+   - "**"
+  tags:
+   - "v*.*.*"
+ schedule:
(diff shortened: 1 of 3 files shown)


jakep36:
This looks good, but let's check with the team on if we want to call this v1.0.0 in #fluree-core. I think db is on 3.X and I don't have a preference on server version numbers, but @aaj3f and @zonotope may have a preference.

jakep36 (approved):

PR #136, opened 2025-04-25
Fix/datafy error

zonotope:
This patch makes consensus messages indicating errors more general by allowing them to be transmitted across the network. It removes the literal exception object from the message and makes receivers reconstruct the exception as necessary from the data within the message.


dpetran (approved):
:fire_extinguisher:

PR #139, opened 2025-05-19
Set tracking response headers

zonotope:
This patch removes the middleware to set response headers in favor of a function that each handler can call because each response between history, transaction, and queries have different shapes which shouldn't be handled by a general middleware. It also adds fuel, policy, and response time values to the transaction completed events so that these values can be returned in the headers, and encodes the policy header as a base64 encoded json string.

### DIFF
--- a/src/fluree/server/handlers/shared.clj
+++ b/src/fluree/server/handlers/shared.clj
@@ -1,6 +1,10 @@
 (ns fluree.server.handlers.shared
-  (:require [fluree.db.util.core :as util])
-  (:import (clojure.lang ExceptionInfo)))
+  (:require [fluree.db.util.core :as util]
+            [fluree.db.util.json :as json])
+  (:import (clojure.lang ExceptionInfo)
+           (java.util Base64)))
+
+(set! *warn-on-reflection* true)
 
 (defn deref!
   "Derefs promise p and throws if the result is an exception, returns it otherwise."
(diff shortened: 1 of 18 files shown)


src/fluree/server/handlers/shared.clj
```diff
@@ -27,8 +31,38 @@
...
+  [response fuel]
+  (with-header response "x-fdb-fuel" (str fuel)))
+
+(defn base64-encode
+  [^String s]
+  (.encodeToString (Base64/getEncoder) (.getBytes s)))
+
+(defn with-policy-header
+  [response policy]
+  (let [encoded-policy (-> policy json/stringify base64-encode)]
```
dpetran, line 60: Is it worth base64 encoding this? It makes the response larger and forces the end user to decode it properly. I think it would be simpler to just use `json/stringify` without b64 encoding it.

test/fluree/server/integration/basic_query_test.clj
```diff
@@ -1,17 +1,17 @@
...
-            [jsonista.core :as json]))
+             :refer [api-post create-rand-ledger json-headers run-test-server]]))
 
 (use-fixtures :once run-test-server)
 
 (deftest ^:integration ^:json query-json-test
   (testing "can query a basic entity w/ JSON"
     (let [ledger-name (create-rand-ledger "query-endpoint-basic-entity-test")
           txn-req     {:body
-                       (json/write-value-as-string
```
dpetran, line 14: Is there a reason we're switching away from [jsonista](https://github.com/metosin/jsonista) here? It's [faster](https://www.metosin.fi/blog/faster-json-processing-with-jsonista) than [cheshire](https://github.com/dakrone/cheshire), which is what we're wrapping in `fluree.db.util.json`.

src/fluree/server/handlers/shared.clj
zonotope: base64 encoding this is the most reliable way to ensure consumers can reliably reconstruct the data. 

http allows multiple headers with the same key, and `x-fdb-policy: foo`, `x-fdb-policy: bar`, and `x-fdb-policy: baz` is interpreted as `x-fdb-policy: foo,bar,baz`, and vice versa. 

i don't want commas to appear in a header payload because commas have special meaning in a header. that's why stringified json isn't enough.

test/fluree/server/integration/basic_query_test.clj
zonotope: we don't have any explicit dependency on jsonista; we only implicitly bring it in because we have a dependency on malli. furthermore, we already have our own json parsing/encoding namespace, so we should always use that. if jsonista is truly better than cheshire, then we should explicitly depend on it and change the implementation of our json encoding/decoding functions to use it. that way we reap those benefits everywhere.

dpetran (approved):
:tractor:

PR #146, opened 2025-06-17
Feature/drop

dpetran:
Add a `/fluree/drop` endpoint to enable the deletion of a ledger.

### DIFF
--- a/src/fluree/server/consensus.clj
+++ b/src/fluree/server/consensus.clj
@@ -9,7 +9,8 @@
 
 (defprotocol Transactor
   (-queue-new-ledger [transactor new-ledger-params])
-  (-queue-new-transaction [transactor new-tx-params]))
+  (-queue-new-transaction [transactor new-tx-params])
+  (-queue-drop-ledger [transactor drop-ledger-params]))
 
 (defn queue-new-ledger
   [transactor ledger-id tx-id txn opts]
@@ -22,3 +23,8 @@
   (log/trace "queue-new-transaction:" txn)
   (let [event-params (events/commit-transaction ledger-id tx-id txn opts)]
     (-queue-new-transaction transactor event-params)))
+
+(defn queue-drop-ledger
+  [transactor ledger-id]
+  (let [event-params (events/drop-ledger ledger-id)]
+    (-queue-drop-ledger transactor event-params)))
--- a/src/fluree/server/consensus/events.clj
+++ b/src/fluree/server/consensus/events.clj
@@ -47,6 +47,13 @@
              :instant   (System/currentTimeMillis)}]
     (with-txn evt txn)))
 
+(defn drop-ledger
+  "Create a new event message to drop an existing ledger."
+  [ledger-id]
+  {:type      :ledger-drop
+   :ledger-id ledger-id
+   :instant   (System/currentTimeMillis)})
+
 (defn commit-transaction
   "Create a new event message to commit a new transaction. The `txn` argument may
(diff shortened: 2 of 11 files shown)


src/fluree/server/consensus/events.clj
```diff
@@ -173,6 +180,24 @@
...
 
+(defn ledger-dropped
+  ([ledger-id _drop-result]
+   {:type :ledger-dropped
+    :ledger-id ledger-id})
+  ([processing-server ledger-id drop-result]
+   (-> (ledger-dropped ledger-id drop-result)
+       (assoc :server processing-server))))
+
+(defn drop-error
```
zonotope, line 191: Why does there need to be a separate error event for drop?

src/fluree/server/handlers/drop.clj
```diff
@@ -0,0 +1,59 @@
...
+        (log/warn drop-resp "Error dropping ledger")
+        (let [error-event (events/drop-error ledger-id drop-resp)]
+          (watcher/deliver-event watcher ledger-id error-event))))))
+
+(defn queue-consensus
+  [consensus watcher ledger-id]
+  (let [drop-resp-ch (consensus/queue-drop-ledger consensus ledger-id)]
+    (monitor-consensus watcher ledger-id drop-resp-ch)))
+
+(defn monitor-watch
```
zonotope, line 26: I think we could generalize the functions in the tranact handler instead of writing new ones. `monitor-consensus` and `monitor-watch` look to be essentially the same as functions in there.

src/fluree/server/consensus.clj
```diff
@@ -22,3 +26,20 @@
   (log/trace "queue-new-transaction:" txn)
   (let [event-params (events/commit-transaction ledger-id tx-id txn opts)]
     (-queue-new-transaction transactor event-params)))
+
+(defn queue-drop-ledger
+  [transactor ledger-id]
+  (let [event-params (events/drop-ledger ledger-id)]
+    (-queue-drop-ledger transactor event-params)))
+
+(defn monitor-consensus
```
zonotope, line 35: I don't think this function should go here. The consensus component should not be concerned with putting messages on the watcher. That's why this function was within the handler namespace hierarchy to begin with because that's the integration point that ties the independent components together. If you're worried about requiring one handler in another, then add it to the shared handler namespace.

src/fluree/server/consensus/raft.clj
```diff
@@ -327,6 +327,8 @@
   consensus/Transactor
   (-queue-new-ledger [group ledger-msg]
     (new-entry-async group ledger-msg))
+  (-queue-drop-ledger [_group ledger-msg]
+    (throw (ex-info "Drop not supported for Raft consensus." (merge ledger-msg {:error :not-implemented}))))
```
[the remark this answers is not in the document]
dpetran: I've updated the error message, but I've started working on getting raft working. I think that will be a separate PR, though.

src/fluree/server/handlers/shared.clj
```diff
@@ -66,3 +70,15 @@
     time   (with-time-header time)
     fuel   (with-fuel-header fuel)
     policy (with-policy-header policy)))
+
+(defn monitor-consensus
```
zonotope, line 74: I think the old name of this function was better. This function doesn't monitor the entire consensus process; it monitors that the event was persisted by the consensus layer to it's own storage/memory before the consensus layer begins processing the event

zonotope (approved):

PR #147, opened 2025-06-26
update save-txn! calls to handle new signature

dpetran:
There was further refactoring and this got missed in the update.

Depends on https://github.com/fluree/db/pull/1041


bplatz (approved):
Great!

PR #148, opened 2025-07-01
Add config into to README

bplatz:
Updates README. The current version was incorrect and old. When the changes for the JSON-LD configuration were introduced, it wasn't updated and nobody would be able to use it as it stood.

Update the startup requirements along with updating docs a little bit.


dpetran (approved):
:spiral_notepad:

PR #150, opened 2025-07-02
Add config profile support

bplatz:
This adds support for profile overrides within the JSON-LD config files.

There is also more logging at startup so you can see what configs are actually set.

The :run-dev deps.edn alias is fixed so it works with new startup

Old profile parsing code no longer used is removed.


src/fluree/server/system.clj
```diff
@@ -126,24 +134,70 @@
...
+        closed-mode (when http-config
+                      (conn-config/get-first-boolean http-config server-vocab/closed-mode))]
+    (log/info "Server configuration summary:")
+    (log/info "  HTTP port:" (or http-port "Not configured"))
+    (log/info "  Consensus mode:" (or consensus-type "Not configured"))
+    (when max-pending-txns
+      (log/info "  Max pending transactions:" max-pending-txns))
+    (log/info "  Storage:" (if storage-path
+                             (str "File-based at " storage-path)
+                             "In-memory"))
```
zonotope, line 166: We theoretically support more storage types than file and memory like s3 and ipfs for example. Whether the other ones actually work or not is another question.

zonotope (approved):
Thanks for doing this!

PR #153, opened 2025-07-02
add /update, /insert, /upsert api endpoints

dpetran:
Add support for the new apis:
- `/update` - same as `/transact`
- `/insert` - takes a plain rdf payload (JSON-LD or Turtle) and inserts with `:insert` semantics`
- `/upsert` - takes a plain rdf payload (JSON-LD or Turtle) and upserts it, replacing existing properties on each subject with the new object.

Depends on https://github.com/fluree/db/pull/1045

### DIFF
--- a/src/fluree/server/consensus/events.clj
+++ b/src/fluree/server/consensus/events.clj
@@ -1,7 +1,8 @@
 (ns fluree.server.consensus.events
   "Common namespace for defining consensus event messages shared across consensus
   protocols"
-  (:require [fluree.db.connection :as connection]
+  (:require [clojure.string :as str]
+            [fluree.db.connection :as connection]
             [fluree.db.track :as-alias track]
             [fluree.db.transact :as transact]
             [fluree.db.util.async :refer [<? go-try]]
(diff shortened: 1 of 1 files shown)


deps.edn
```diff
@@ -48,7 +48,8 @@
    :extra-deps  {org.clojure/tools.namespace {:mvn/version "1.5.0"}
                  clj-http/clj-http           {:mvn/version "3.13.0"}
                  criterium/criterium         {:mvn/version "0.4.6"}
-                 integrant/repl              {:mvn/version "0.3.3"}}}
+                 integrant/repl              {:mvn/version "0.3.3"}}
+   :jvm-opts    ["-Djdk.attach.allowAttachSelf"]}
```
zonotope, line 52: What is this for?

src/fluree/server/consensus/events.clj
```diff
@@ -24,13 +24,16 @@
   (map? x))
 
 (defn with-txn
-  [evt txn]
-  (cond (txn-address? txn)
-        (assoc evt :txn-address txn)
+  [evt txn {:keys [format] :as _opts}]
```
zonotope, line 27: I don't think we should be pushing opts further into the call stack. If anything, we should be pushing them up further towards the api boundary.

I think this should be two different functions, one for turtle and one for json. One could call the other to avoid repeated code.

deps.edn
dpetran: This allows me to cancel a running process from the repl in emacs - I accidentally hung the whole jvm while adding drop support and this was how I was able to debug it. I didn't intend to add it to git, but I don't think it does any harm and is useful in general for development. I'm fine removing it, but if you're fine then I wouldn't mind keeping it around.

dpetran:
Rebased on main.

src/fluree/server/consensus/events.clj
```diff
@@ -64,7 +81,9 @@
              :ledger-id ledger-id
              :opts      opts
              :instant   (System/currentTimeMillis)}]
-    (with-txn evt txn)))
+    (if (= :turtle (:format opts))
+      (with-turtle-txn evt txn)
+      (with-txn evt txn))))
```
zonotope, line 86: this is minor, but perhaps this function could have a more specific name to further disambiguate it from turtle.

zonotope (approved):
🥼

PR #154, opened 2025-07-02
Update server to support drop on nexus-server

dpetran:
Getting `drop` to work on `nexus-server` required some additional tweaks.

Depends on https://github.com/fluree/db/pull/1055

### DIFF
--- a/src/fluree/server/consensus/events.clj
+++ b/src/fluree/server/consensus/events.clj
@@ -231,3 +231,10 @@
       (ledger-created? evt)
       (ledger-dropped? evt)
       (error? evt)))
+
+(defn watcher-id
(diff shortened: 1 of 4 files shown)


src/fluree/server/consensus/events.clj
```diff
@@ -225,4 +229,12 @@
   [evt]
   (or (transaction-committed? evt)
       (ledger-created? evt)
+      (ledger-dropped? evt)
       (error? evt)))
+
+(defn watcher-id
```
zonotope, line 235: This is minor, but I think we should update our current code to use this new function instead of using the tx-id directly.

zonotope (approved):
👾

PR #157, opened 2025-07-03
Add tests for ConfigurationValue substitution

bplatz:
## Summary
Fluree Server relies on environment-based configuration through ConfigurationValue, but there were no tests or documentation for this functionality. This PR addresses that gap by:

- Adding comprehensive test suite for ConfigurationValue substitution functionality
- Tests cover Java property substitution, environment variables, default values, and priority ordering
- Creating helper functions to extract resolved configuration values from the system map
- Adding documentation explaining how to use ConfigurationValue for dynamic configuration

## Changes
1. **New test file**: `test/fluree/server/config_substitution_test.clj`
   - 6 test cases covering all ConfigurationValue scenarios
   - Helper functions for extracting resolved values from system map
   - Test fixture to prevent test pollution from Java system properties

2. **Updated README.md**:
   - Documented ConfigurationValue usage with examples
   - Explained resolution priority (Java property > env var > default)
   - Added practical examples for HTTP port, cache size, and storage path
   - Fixed line wrapping to stay within 80 characters

## Test plan
- [x] Run tests with `clojure -X:test`
- [x] Verify all 6 test cases pass
- [x] Tests handle both string and numeric value types correctly
- [x] Tests clean up Java system properties to prevent test pollution
- [x] Documentation is clear and includes working examples

### DIFF
--- a/resources/file-config.jsonld
+++ b/resources/file-config.jsonld
@@ -52,7 +52,7 @@
     {
       "@id": "http",
       "@type": "API",
-      "httpPort": 8090,
+      "httpPort": 58090,
       "maxTxnWaitMs": 120000
     }
   ],
--- a/test/fluree/server/config_substitution_test.clj
+++ b/test/fluree/server/config_substitution_test.clj
@@ -1,55 +1,27 @@
 (ns fluree.server.config-substitution-test
   (:require [clojure.test :refer [deftest is testing use-fixtures]]
+            [fluree.db.connection.system :as conn-system]
             [fluree.server.system :as system]))
 
+(def test-properties
+  ["fluree.test.cache.size"
+   "fluree.test.parallelism"
+   "fluree.test.required.prop"])
+
 (defn clear-test-properties
-  "Test fixture that ensures Java system properties used in tests are cleared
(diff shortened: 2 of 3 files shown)


bplatz:
@zonotope I've taken a slightly different tact here and just focused on the configuration values being picked up and not on a thorough test or relying on the internals of the returned "system". The goal was really to have some test that would break if fluree/db changed how it handles configs, since fluree/server is heavily dependent on them and in particular is dependent on substituting variables through java props for env, which fluree/db likely relies much less on.

This overall approach is pretty simple, but at least we have some tests now where before we had none.

zonotope:
> @zonotope I've taken a slightly different tact here and just focused on the configuration values being picked up and not on a thorough test or relying on the internals of the returned "system". The goal was really to have some test that would break if fluree/db changed how it handles configs, since fluree/server is heavily dependent on them and in particular is dependent on substituting variables through java props for env, which fluree/db likely relies much less on.
> 
> This overall approach is pretty simple, but at least we have some tests now where before we had none.

This works for me, and the tests look good.

PR #158, opened 2025-07-07
Fix: /update endpoint properly handles ledger from header

bplatz:
## Summary
- Fixed a bug where the `/update` endpoint was not properly handling the ledger parameter when provided via the `fluree-ledger` header
- The endpoint now adds the ledger to the transaction body before passing it to the underlying fluree.db library

## Problem
The `/update` endpoint would fail with "Invalid transaction, missing required key: ledger" when the ledger was only provided in the header, even though the handler code was designed to accept it from either the header or the request body.

## Solution
Modified the update handler to check if the ledger was provided via header but not in the transaction body, and if so, adds it to the transaction before passing it to the underlying library.

## Test plan
- [x] Verified that the README examples test now passes with ledger in header
- [x] Confirmed that providing ledger in the body still works as before
- [x] Ensured consistency with how `/insert` and `/upsert` endpoints handle the ledger parameter

Note: A subsequent PR will include updated README documentation and tests that demonstrate the proper usage of the new transaction endpoints with headers.

### DIFF
--- a/src/fluree/server/handlers/transact.clj
+++ b/src/fluree/server/handlers/transact.clj
@@ -54,11 +54,16 @@
   (let [txn       (fluree/format-txn body opts)
         ledger-id (or (:ledger opts)
                       (extract-ledger-id txn))
+        ;; Ensure the transaction includes the ledger for the underlying update! function
+        txn-with-ledger (if (and (:ledger opts)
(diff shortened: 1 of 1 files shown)


dpetran:
Do the tests actually test the `/update` endpoint with and without the `fluree-ledger` header? Or is the "Test plan" still TODO?

PR #162, opened 2025-07-09
update json-ld dep

dpetran:
The git dep on json-ld conflicts with the mvn dep on json-ld in the db library, this ensures that when server is used as a library there is no transitive dependency conflict (git deps and maven deps are not comparable by clojure, so the build fails).


zonotope (approved):
🅱️

PR #163, opened 2025-07-11
Enable ledger creation without initial transaction data

bplatz:
## Summary
- Allow t=0 for genesis commits in API schema validation
- Extract hash from commit ID for consistent response format across all ledger creation scenarios
- Add shared functions for genesis commit handling between standalone and Raft consensus
- Fix broadcast action to "new-commit" for client compatibility with remote connections
- Update documentation to show simple ledger creation as primary example
- Add test coverage for creation without initial data
- Ensure README examples and tests are consistent with new transaction sequence

## Test plan
- [x] Add unit test for creating ledger without initial data
- [x] Update existing integration tests to handle new transaction numbering
- [x] Verify API schema accepts t=0 genesis commits
- [x] Confirm broadcast compatibility with remote connections
- [x] Update README examples and ensure consistency with test suite
- [x] Run cljfmt fix to ensure code formatting

### DIFF
--- a/src/fluree/server/consensus/shared/create.clj
+++ b/src/fluree/server/consensus/shared/create.clj
@@ -1,0 +1,42 @@
+(ns fluree.server.consensus.shared.create
+  "Shared functions for ledger creation across consensus mechanisms."
+  (:require [fluree.db.api :as fluree]
+            [fluree.db.util.log :as log]))
+
+(defn hash-from-id
+  "Extracts the hash portion from a commit ID.
+  Commit IDs have format: 'fluree:commit:sha256:b<base32-hash>'"
+  [commit-id]
+  (when (string? commit-id)
+    (let [prefix "fluree:commit:sha256:b"]
+      (when (.startsWith ^String commit-id prefix)
+        (subs commit-id (count prefix))))))
+
+(defn genesis-result
+  "Creates a commit result map for a genesis commit."
+  [ledger]
+  (let [db     (fluree/db ledger)
+        commit (:commit db)]
+    {:db      db
+     :address (:address commit)
+     :hash    (or (:hash commit)
+                  (hash-from-id (:id commit)))}))
+
+(defn file-result
+  "Creates a commit result map with file metadata structure.
+  Used by Raft consensus. Genesis commits have data files but with empty assertions."
+  [ledger]
+  (let [db         (fluree/db ledger)
+        commit     (:commit db)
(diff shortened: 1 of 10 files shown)


src/fluree/server/handlers/create.clj
```diff
@@ -23,20 +23,32 @@
...
-                         (srv-tx/extract-ledger-id txn))
-        opts*        (dissoc opts :identity) ; Remove identity option because the
-                                             ; request should have been validated
-                                             ; upstream, and there are no policies
-                                             ; in an empty ledger to allow any
-                                             ; actions
-        commit-event (deref! (create-ledger! consensus watcher ledger-id txn opts*))
-
-        body (srv-tx/commit-event->response-body commit-event)]
+  (let [has-txn-data? (contains? body "insert")
```
zonotope, line 43: this is minor, but i think this reads a little better as a function called like 

```clojure
(when (has-txn-data? body)
  ...
```

src/fluree/server/consensus/shared/create.clj
```diff
@@ -0,0 +1,42 @@
...
+  (:require [fluree.db.api :as fluree]
+            [fluree.db.util.log :as log]))
+
+(defn hash-from-id
+  "Extracts the hash portion from a commit ID.
+  Commit IDs have format: 'fluree:commit:sha256:b<base32-hash>'"
+  [commit-id]
+  (when (string? commit-id)
+    (let [prefix "fluree:commit:sha256:b"]
+      (when (.startsWith ^String commit-id prefix)
```
zonotope, line 12: you can use `clojure.string/starts-with?` here.

src/fluree/server/consensus/shared/create.clj
```diff
@@ -0,0 +1,42 @@
+(ns fluree.server.consensus.shared.create
+  "Shared functions for ledger creation across consensus mechanisms."
+  (:require [fluree.db.api :as fluree]
+            [fluree.db.util.log :as log]))
+
+(defn hash-from-id
+  "Extracts the hash portion from a commit ID.
+  Commit IDs have format: 'fluree:commit:sha256:b<base32-hash>'"
+  [commit-id]
+  (when (string? commit-id)
```
zonotope, line 10: It might be useful to break all of this out into a `commit-id?` function. That function would probably come in handy and be reused, and you could just do `(when (commit-id? commit-id) ...`

PR #164, opened 2025-07-16
rename util ns and update db

dpetran:
Depends on https://github.com/fluree/db/pull/1068


bplatz (approved):
Awesome, please merge! ⛑️

PR #165, opened 2025-07-29
update crypto api usages

dpetran:
The default keypair generation now produces did:key strings as the :id key. I converted the existing ones by taking the private key and calling `fluree.db.did/private->did-map` with it.

I also updated calls `crypto/create-jws`, which with the new signing algorithm cannot derive the pubkey from the payload+sig, so needs to have the pubkey in the JWS header.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -6,7 +6,7 @@
 
         integrant/integrant {:mvn/version "0.10.0"}
         com.fluree/crypto   {:git/url "https://github.com/fluree/fluree.crypto.git"
-                             :git/sha "cd92370b7c49aac78720833f520787dcebe86b9e"}
+                             :git/sha "065c425d98c97cb648f01ad3c708f2c71c80e11d"}
 
         ;; network, consensus
         com.github.fluree/raft {:git/tag "v1.0.0-beta2"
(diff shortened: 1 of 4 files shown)


bplatz:
> After updating this, I think we should change the `crypto/create-jws` function to include the pubkey by default, otherwise we can't verify the JWS we create, which is what I think it should do out of the box.

I tend to agree. There is definitely a use case for not including it (you've supplied it out of band to select parties and don't want to disclose it) but I think for our use case where we want the verification as much for 3rd party verification as anything else then we'd generally want the public key as part of the message.

Do you think we should just not even support a no public key option in our library, or create a different API so we have one API for having a public key and other for not having it? Thoughts?

dpetran:
> Do you think we should just not even support a no public key option in our library, or create a different API so we have one API for having a public key and other for not having it? Thoughts?

I think we should support no pubkey, the reasons you mentioned above seem reasonable. Here's the api that makes sense to me:

`(crypto/create-jws payload signing-key)` -> derives pubkey and adds it to header
`(crypto/create-jws payload signing-key-map)` -> uses the pubkey from the signing-key map
`(crypto/create-jws payload signing-key {:include-pubkey false})` -> does not add pubkey to JWS header.

Basically, just flipping the default behavior of `:include-pubkey`.

I also think we should consider having verify throw an exception if something doesn't work, rather than catching the exception and returning it as a value.

zonotope (approved):
🤝

PR #179, opened 2025-09-18
Update to latest fluree/db

bplatz:
Update to latest fluree/db dependency, fixes one deprecated function call. Based on https://github.com/fluree/server/pull/178


dpetran (approved):
:cactus:

PR #180, opened 2025-09-20
Feature/reindex command

dpetran:
This PR adds the ability to reindex ledgers via a command line interface.

## Single ledger
```
java -jar server-<version>.jar --reindex myledger
```

## All ledgers
```
java -jar server-<version>.jar --reindex --all
```

Reindexing is triggered in parallel and the server will exit once all the jobs have finished or failed, with logging indicating each ledger's status.

### DIFF
--- a/src/fluree/server/system.clj
+++ b/src/fluree/server/system.clj
@@ -213,7 +213,8 @@
     (when reindex
       (let [connection-key (first (filter #(isa? % :fluree.db/connection) (keys system)))]
         (when-let [conn (get system connection-key)]
-          (async/<!! (task/reindex conn reindex)))))
+          (let [{:keys [status]} (async/<!! (task/reindex conn reindex))]
+            (System/exit status)))))
     system))
 
 (defn start-file
--- a/src/fluree/server/system.clj
+++ b/src/fluree/server/system.clj
@@ -198,39 +198,34 @@
       (log/info "  Closed mode: enabled"))))
 
 (defn start-config
-  ([config]
-   (start-config config nil))
-  ([config profile]
-   (let [json-config (if (string? config)
-                       (json/parse config false)
-                       config)
-         config-with-profile (if profile
-                               (config/apply-profile json-config profile)
-                               json-config)
-         parsed-config (config/parse config-with-profile)]
-     (log-config-summary parsed-config)
-     (conn-system/initialize parsed-config))))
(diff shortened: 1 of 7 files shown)


src/fluree/server/task.clj
```diff
@@ -0,0 +1,55 @@
+(ns fluree.server.task
```
zonotope, line 1: This namespace should be called `fluree.server.reindex`

src/fluree/server/system.clj
```diff
@@ -198,39 +200,40 @@
...
+  (let [json-config         (if (string? config)
+                              (json/parse config false)
+                              config)
+        config-with-profile (if profile
+                              (config/apply-profile json-config profile)
+                              json-config)
+        parsed-config       (config/parse config-with-profile)
+        system              (conn-system/initialize parsed-config)]
+    (log-config-summary parsed-config)
+    (when reindex
```
zonotope, line 213: This should be an `if`, with the else branch returning the system.

src/fluree/server/system.clj
```diff
@@ -198,39 +200,40 @@
...
+      (let [connection-key (first (filter #(isa? % :fluree.db/connection) (keys system)))]
+        (when-let [conn (get system connection-key)]
```
zonotope, line 215: this should be an `if-let`, with the else branch exiting with an error about no connection being found.

src/fluree/server/task.clj
```diff
@@ -0,0 +1,55 @@
...
+    (let [records    (<? (nameservice/all-records primary-publisher))
+          aliases    (extract-ledger-aliases records)
+          in-ch      (async/to-chan! aliases)
+          out-ch     (async/chan parallelism (remove #(zero? (:status %))))
+          reindex-fn (fn [alias ch]
+                       (-> (reindex-single conn alias)
+                           (async/pipe ch)))]
+      (log/info "Found" (count aliases) "ledgers to reindex:" aliases)
+      (async/pipeline-async parallelism out-ch reindex-fn in-ch)
+      (let [failed (<? (async/into [] out-ch))]
```
zonotope, line 43: `async/into` will never have an exception on the channel it returns, so you should use `<!` instead.

zonotope (approved):
There are a few changes I'd like to see, but it looks good otherwise.

PR #184, opened 2025-10-09
Support LOG_LEVEL environment variable

bplatz:
## Summary

Adds `LOG_LEVEL` environment variable support to logback configuration files for runtime log level control.

## Changes

- Added `LOG_LEVEL` property with `INFO` default fallback
- Applied to all logger levels (root, fluree.server, fluree.db, fluree.db.consensus, fluree.db.messaging)
- Updated both `dev/logback.xml` and `resources/logback.xml`

## Usage

```bash
# Set log level via environment variable
export LOG_LEVEL=DEBUG
java -jar target/server-*.jar

# Or inline
LOG_LEVEL=TRACE java -jar target/server-*.jar
```

Valid levels: `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`

## Backward Compatibility

✅ Defaults to `INFO` if `LOG_LEVEL` not set - no behavior change for existing deployments

### DIFF
--- a/dev/logback.xml
+++ b/dev/logback.xml
@@ -1,5 +1,6 @@
 <!-- Logback configuration. See http://logback.qos.ch/manual/index.html -->
 <configuration scan="true" scanPeriod="10 seconds">
+    <property scope="context" name="LOG_LEVEL" value="${LOG_LEVEL:-INFO}"/>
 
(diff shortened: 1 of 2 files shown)


zonotope (approved):
🗝️

PR #202, opened 2026-01-26
make otel tracing self contained

dpetran:
Add java instrumentation agent to the repo so developers don't need to acquire the jar file themselves and update the deps.edn file with the local path.

This should make it easier to use instrumentation during development.


zonotope (approved):
🐙

